PR #13, opened 2018-08-16
Update README.md 

foldingbeauty:
tell docker to run an interactive and add a tty


danielsagi:
see https://github.com/aquasecurity/kube-hunter/pull/14

PR #25, opened 2018-08-21
print() is a function in Python 3

cclauss:
Just like #15 and #19 but in a new file this time.  Perhaps it is time to add some automated testing like Travis CI, Circle CI, Appveyor, etc.  https://github.com/marketplace/category/continuous-integration


lizrice:
Yes we definitely should add some automated tests! I'll open an issue

PR #50, opened 2018-10-07
Add more logs

oriagmon:
Added logs for most of the tests and discoveries we are doing
<Passive/ Active hunter is attempting to ....>
<Attempting to discover .... >

### DIFF
--- a/src/modules/discovery/hosts.py
+++ b/src/modules/discovery/hosts.py
@@ -73,6 +73,7 @@ class HostDiscovery(Hunter):
 
     def get_cloud(self, host):
         try:
+            logging.debug("Passive hunter is attempting to access azure's cloud")
             metadata = requests.get("http://www.azurespeed.com/api/region?ipOrUrl={ip}".format(ip=host)).text
         except requests.ConnectionError as e:
             logging.info("- unable to check cloud: {0}".format(e))
@@ -82,6 +83,7 @@ class HostDiscovery(Hunter):
 
     def is_azure_pod(self):
         try:
+            logging.debug("Passive hunter is attempting to access azure's pod")
             if requests.get("http://169.254.169.254/metadata/instance?api-version=2017-08-01", headers={"Metadata":"true"}, timeout=5).status_code == 200:
                 return True
         except requests.exceptions.ConnectionError:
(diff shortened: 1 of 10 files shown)


src/modules/hunting/dashboard.py
```diff
@@ -23,6 +23,7 @@ def __init__(self, event):
         self.event = event
 
     def get_nodes(self):
+        logging.debug("Active hunter is attempting to get nodes types of the cluster")
```
lizrice: Not quite English - how about "Active hunter is attempting to get node information"?

src/modules/discovery/ports.py
```diff
@@ -20,9 +20,10 @@ def __init__(self, event):
         self.port = event.port
 
     def execute(self):
-        logging.debug("host {0} try ports {1}".format(self.host, default_ports))
+        logging.debug("host {0} try reach ports: {1}".format(self.host, default_ports))
```
lizrice: Not quite English - either "try ports" or "try to reach ports" (I think I prefer "try ports" as it's more succinct)

src/modules/discovery/hosts.py
```diff
@@ -73,6 +73,7 @@ def execute(self):
 
     def get_cloud(self, host):
         try:
+            logging.debug("Passive hunter is attempting to access azure's cloud")
```
lizrice: I think this is really more like "checking whether this is deployed in Azure"

src/modules/discovery/hosts.py
```diff
@@ -82,6 +83,7 @@ def get_cloud(self, host):
 
     def is_azure_pod(self):
         try:
+            logging.debug("Passive hunter is attempting to access azure's pod")
```
lizrice: "Attempting to access Azure Metadata API"

src/modules/discovery/hosts.py
```diff
@@ -110,7 +113,8 @@ def azure_metadata_discovery(self):
 
     # for normal scanning
     def scan_interfaces(self):
-        try: 
+        try:
+            logging.debug("Passive hunter is attempting to scan interfaces")
```
lizrice: "attempting to get external IP address"

src/modules/discovery/kubelet.py
```diff
@@ -53,6 +55,7 @@ def get_secure_access(self):
             self.publish_event(SecureKubeletEvent(secure=True, anonymous_auth=False))
 
     def ping_kubelet(self):
+        logging.debug("Attempting to ping kubelet")
```
lizrice: I know the function is called ping_kubelet but I think that's a bit misleading ("ping" to me implies ICMP rather than HTTP). Maybe "Attempting to get pod info from kubelet"?

oriagmon:
Great! thanks Liz :-)

PR #52, opened 2018-10-07
Etcd hunter

oriagmon:
I believe this branch is finished! :-) 
It checks for these passive hunters:
* Remote read access
* Remote version disclosure
* Remote unauthorized access (no certificate & over HTTP & no authentication)
And for the active hunter:
* Tries to gain a remote write access.
----------------------------------------------
It was tested and worked as expected on my machine using both vulnerable cluster and non vulnerable cluster.

### DIFF
--- a/src/modules/discovery/etcd.py
+++ b/src/modules/discovery/etcd.py
@@ -23,6 +23,11 @@ class etcdRemoteVersionDisclosureEvent(Service, Event):
     """Remote version disclosure might give an attacker a valuable data to attack a cluster"""
     def __init__(self):
         Service.__init__(self, name="Etcd Remote version disclosure")
+class etcdAccessEnabledWithoutAuthEvent(Service, Event):
+    """Remote version disclosure might give an attacker a valuable data to attack a cluster"""
+    def __init__(self):
+        Service.__init__(self, name="Etcd is accessible without authorization")
+
 
 
 @handler.subscribe(OpenPortEvent, predicate= lambda p: p.port == 2379)
@@ -76,6 +81,7 @@ class etcdRemoteAccess(Hunter):
 
     def execute(self):
         if (self.version_disclosure()):
+           self.publish_event(etcdAccessEnabledWithoutAuthEvent())#if version is accessible we can publish "no auth event".
            self.db_keys_disclosure()
            self.db_keys_write_access()
 
--- a/src/modules/discovery/etcd.py
+++ b/src/modules/discovery/etcd.py
@@ -7,11 +7,9 @@ from ...core.events import handler
 from ...core.events.types import Event, OpenPortEvent, Service
 from ...core.types import Hunter
 
-"""Etcd is a DB that stores cluster's data,
-    it contains configuration and current state information, and might contain secrets"""
-#services:
+# Service:
 
-class etcdAccessEvent(Service, Event):
+class EtcdAccessEvent(Service, Event):
     """Etcd is a DB that stores cluster's data, it contains configuration and current state information, and might contain secrets"""
     def __init__(self):
         Service.__init__(self, name="Etcd")
@@ -19,7 +17,7 @@ class etcdAccessEvent(Service, Event):
 
 
 @handler.subscribe(OpenPortEvent, predicate= lambda p: p.port == 2379)
-class etcdRemoteAccess(Hunter):
+class EtcdRemoteAccess(Hunter):
     """Etcd Remote Access
     Checks for remote availability of etcd, version, read access, write access
     """
@@ -28,7 +26,7 @@ class etcdRemoteAccess(Hunter):
 
     def execute(self):
         try:
-            self.publish_event(etcdAccessEvent())
+            self.publish_event(EtcdAccessEvent())
         except Exception as e:
             import traceback
             traceback.print_exc()
--- a/src/modules/discovery/etcd.py
+++ b/src/modules/discovery/etcd.py
@@ -11,22 +11,10 @@ from ...core.types import Hunter
     it contains configuration and current state information, and might contain secrets"""
 #services:
 
-class etcdRemoteWriteAccessEvent(Service, Event):
-    """Remote write access might grant an attacker full control over the kubernetes cluster"""
+class etcdAccessEvent(Service, Event):
+    """Etcd is a DB that stores cluster's data, it contains configuration and current state information, and might contain secrets"""
     def __init__(self):
-        Service.__init__(self, name="Etcd Remote Write Access Event")
-class etcdRemoteReadAccessEvent(Service, Event):
-    """Remote read access might expose to an attacker cluster's possible exploits, secrets and more."""
-    def __init__(self):
-        Service.__init__(self, name="Etcd Remote Read Access Event")
-class etcdRemoteVersionDisclosureEvent(Service, Event):
-    """Remote version disclosure might give an attacker a valuable data to attack a cluster"""
-    def __init__(self):
-        Service.__init__(self, name="Etcd Remote version disclosure")
-class etcdAccessEnabledWithoutAuthEvent(Service, Event):
-    """Etcd is accessible without authorization, it would allow a potential attacker to gain access to the etcd"""
-    def __init__(self):
-        Service.__init__(self, name="Etcd is accessible without authorization")
+        Service.__init__(self, name="Etcd Access")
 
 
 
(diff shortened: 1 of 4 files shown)


src/modules/hunting/etcd.py
```diff
@@ -0,0 +1,142 @@
...
+        logging.debug("Passive hunter is attempting to access etcd without authorization")
+        r_not_secure = "http://{host}:{port}/version".format(host=self.event.host, port=2379)
+        res = helperFuncDo2Requests(r_not_secure, r_not_secure)# We dont have to do 2 requests this time
+        if res:
+            self.publish_event(etcdAccessEnabledWithoutAuthEvent(res.content))
+            return True
+        return False
+
+    def execute(self):
+        if self.version_disclosure():
```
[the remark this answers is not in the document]
lizrice: Another thing to consider would be using the version disclosure test to publish two separate events: one for secure access, one for not-secure access. Then you could have something along these lines:

```python
@handler.subscribe(EtcdInsecureAccessEvent)
class EtcdRemoteAccess(Hunter):
    def __init__(self, event):
        self.event = event
        self.protocol = "http"

    def db_keys_disclosure(self):
        r = requests.get("{protocol}://{host}:{port}/v2/keys".format(protocol=self.protocol, host=self.event.host, port=2379))
        self.content = r.content
        return (r.status_code == 200 and r.content != "")

    def execute(self):
        if self.db_keys_disclosure():
             self.publish_event(EtcdRemoteReadAccessEvent(self.content))

@handler.subscribe(EtcdSecureAccessEvent)
class EtcdSecureRemoteAccess(EtcdRemoteAccess):
    def __init__(self, event):
        self.event = event
        self.protocol = "https://"
```

The Insecure version would run if we established that we could access on http://, and the Secure version would run if we established that we could access on https://,  inheriting all the same tests as the Insecure one, but would use the different protocol string. You then wouldn't need the helper functions as the same logic would be taken care of by the classing, which I think is a bit nicer. 

We could also have another prarameter on the events like EtcdRemoteReadAccessEvent to indicate whether it was the secure or insecure version

src/modules/hunting/etcd.py
```diff
@@ -0,0 +1,142 @@
...
+        self.event = event
+
+    def db_keys_write_access(self):
+        logging.debug(self.event.host)
+        logging.debug("Active hunter is attempting to write keys remotely")
+        data = {
+            'value': 'remote write access penetration'
+        }
+        r_secure = "https://{host}:{port}/v2/keys/message".format(host=self.event.host, port=2379)
+        r_not_secure = "https://{host}:{port}/v2/keys/message".format(host=self.event.host, port=2379)
```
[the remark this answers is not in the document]
oriagmon: Interesting thoughts, I will think about it.

src/modules/hunting/etcd.py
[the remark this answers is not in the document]
oriagmon: Are you sure?
Maybe I just prefer the current version just because I don't like splitting the code into different classes and methods; But I think it would make the code harder to understand since we would have 2 classes for the subscribers (instead of 1) that both publishes the same events (since we don't distinguish between an insecure- HTTP  and an secure- HTTPS events).
and 2 more events ( EtcdSecureAccessEvent, EtcdInsecureAccessEvent).

That is true we wouldn't have the Helper functions, but I think they are intuitive and simple to understand.

src/modules/discovery/etcd.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+    def __init__(self, event):
+        self.event = event
+
+    def execute(self):
+        try:
+            self.publish_event(etcdAccessEvent())
+        except Exception as e:
+            import traceback
+            traceback.print_exc()
```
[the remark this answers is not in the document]
oriagmon: Just the exception, forgot to remove it :D

src/modules/hunting/etcd.py
```diff
@@ -0,0 +1,146 @@
...
+from ...core.events import handler
+from ...core.events.types import Vulnerability, Event, OpenPortEvent
+from ...core.types import ActiveHunter, Hunter, KubernetesCluster, InformationDisclosure, RemoteCodeExec, UnauthenticatedAccess, AccessRisk
+
+""" Helper functions """
+
+# Will attempt to do request "req1" with the optional parameters.
+# If fails it will attempt to do "req2" with the optional parameters.
+# If once of the request success this method will return True, if both fail- False.
+def helperFuncDo2Requests(req1, req2, is_verify=False, data=None, req_type="get"):
```
danielsagi: What I saw in your usage of this function:
* You always use the default parameters
* req1 and req2 are actually the same, except for the protocol.

Take a look at this function:
```python
# just send the url without the http/s prefix
def do_request(url):
    try:
        results = map(lambda protocol: requests.get("{}://{}".format(protocol, url), verify=False), ["http", "https"])
        return (result for result in results if result.status_code==200).next()
    except:
        return False
```
I think it is far better to write short and understandable code, rather than a multipurpose helper function, which takes the attention away from the real Hunter. 
Though, looking at what Liz suggested about separating the code to insecure and secure can be somewhat more understandable and effective (less actual requests being sent). But in any case, a point to consider.

src/modules/hunting/etcd.py
```diff
@@ -81,66 +50,57 @@ def db_keys_write_access(self):
...
 
     def unauthorized_access(self):
-        logging.debug(self.event.host)
-        logging.debug("Passive hunter is attempting to access etcd without authorization")
-        r_not_secure = "http://{host}:{port}/version".format(host=self.event.host, port=2379)
-        res = helperFuncDo2Requests(r_not_secure, r_not_secure)  # We don't have to do 2 requests this time
-        if res:
-            return True
-        return False
+        logging.debug(self.event.host + " Passive hunter is attempting to access etcd without authorization")
```
lizrice: The code is fine but the comment and function name are misleading. Lots of people find authentication and authorization confusing so let me try to explain. 
* Here we're making an insecure (http) request, passing no identity information. 
* You can also make a secure (https) request without passing identity information. This encrypts the communication. 
* If you pass identity information (like a bearer token, or a user name & password) the other end might "authenticate" you i.e. check your identity by verifying that the password is correct for the username. There's not a lot of point passing identity information if you're not also encrypting the communication, otherwise anyone could just pick up your identity and pretend to be you, so realistically an authenticated connection will also be encypted (https), but not all https traffic is authenticated. 
* If you are authenticated, the other knows who you are, and might then also make you go through an "authorization" process: "now that I know who you are, I'm going to check whether you have the authority to make this request". 

So in this case it's really just trying to access etcd unencrypted (aka insecure), over http.

src/modules/hunting/etcd.py
```diff
@@ -81,66 +50,57 @@ def db_keys_write_access(self):
...
             'value': 'remotely written data'
         }
-        r_secure = "https://{host}:{port}/v2/keys/message".format(host=self.event.host, port=2379)
-        r_not_secure = "http://{host}:{port}/v2/keys/message".format(host=self.event.host, port=2379)
-        res = helperFuncDo2Requests(r_secure, r_not_secure)
-        if res:
-            self.write_evidence = res.content
-            return True
-        return False
+        r = "{protocol}://{host}:{port}/v2/keys/message".format(host=self.event.host, port=2379, protocol=self.protocol,
```
lizrice: I think you're missing a requests.post() or something?

src/modules/hunting/etcd.py
```diff
@@ -81,66 +50,57 @@ def db_keys_write_access(self):
...
-        res = helperFuncDo2Requests(r_not_secure, r_not_secure)  # We don't have to do 2 requests this time
-        if res:
-            return True
-        return False
+        logging.debug(self.event.host + " Passive hunter is attempting to access etcd without authorization")
+        r = requests.get("http://{host}:{port}/version".format(host=self.event.host, port=2379), verify=False)
+        return r.content if r.status_code == 200 and r.content != '' else False
 
     def execute(self):
+        if self.unauthorized_access():  # decide between http and https protocol
```
lizrice: This approach to trying insecure first and then falling back to https is really neat, I like it!

src/modules/hunting/etcd.py
```diff
@@ -81,66 +50,57 @@ def db_keys_write_access(self):
...
-        logging.debug("Passive hunter is attempting to read etcd keys remotely")
-        r_secure = "https://{host}:{port}/v2/keys".format(host=self.event.host, port=2379)
-        r_not_secure = "http://{host}:{port}/v2/keys".format(host=self.event.host, port=2379)
-        res = helperFuncDo2Requests(r_secure, r_not_secure)
-        if res:
-            self.keys_evidence = res.content
-            return True
-        return False
+        logging.debug(self.event.host + " Passive hunter is attempting to read etcd keys remotely")
+        r = requests.get(
```
lizrice: Is there any chance of this (and other calls to request.get/put) failing? Maybe need a try / except to look for connection errors?

src/modules/hunting/etcd.py
```diff
@@ -81,66 +50,57 @@ def db_keys_write_access(self):
...
+        r = requests.get("http://{host}:{port}/version".format(host=self.event.host, port=2379), verify=False)
+        return r.content if r.status_code == 200 and r.content != '' else False
 
     def execute(self):
+        if self.unauthorized_access():  # decide between http and https protocol
+            self.protocol = 'http'
         if self.version_disclosure():
             self.publish_event(EtcdRemoteVersionDisclosureEvent(self.version_evidence))
-            if self.unauthorized_access():
+            if self.protocol == 'http' and self.unauthorized_access():
```
lizrice: I think you could publish the event after the first time you try self.unauthorized_access() a few lines above (and if it succeeds set http and publish the event at the same time)

src/modules/hunting/etcd.py
oriagmon: Thanks you!

src/modules/hunting/etcd.py
oriagmon: Yes, there is always a chance for this.
We should catch here requests.exceptions.ConnectionError.

src/modules/hunting/etcd.py
oriagmon: The event is indeed logically can be published there, but we won't have the evidence for it.
we can OFC change the unauthorized_access() to save version evidence on success but it would make the method less simple so I have chose to publish it after we have received the evidence.

That being said, I think we can make that logic a bit niter:
`    def execute(self):
        if self.insecure_access():  # make a decision between http and https protocol
            self.protocol = 'http'
        if self.version_disclosure():
            self.publish_event(EtcdRemoteVersionDisclosureEvent(self.version_evidence))
            **if self.protocol == 'http':  # If protocol is HTTP, self.insecure_access() had succeeded so we dont have to do the same request again***
                self.publish_event(EtcdAccessEnabledWithoutAuthEvent(self.version_evidence))
            if self.db_keys_disclosure():
                self.publish_event(EtcdRemoteReadAccessEvent(self.keys_evidence))
`

src/modules/hunting/etcd.py
```diff
@@ -75,31 +82,40 @@ def __init__(self, event):
...
-            "{protocol}://{host}:{port}/v2/keys".format(protocol=self.protocol, host=self.event.host, port=2379),
-            verify=False)
-        self.keys_evidence = r.content if r.status_code == 200 and r.content != '' else False
-        return self.version_evidence
+        try:
+            r = requests.get(
+                "{protocol}://{host}:{port}/v2/keys".format(protocol=self.protocol, host=self.event.host, port=2379),
+                verify=False)
+            self.keys_evidence = r.content if r.status_code == 200 and r.content != '' else False
+            return self.version_evidence
```
lizrice: Should this be self.keys_evidence?

PR #53, opened 2018-10-09
Access secrets hunter

oriagmon:
Attempt to access the secrets under the default pod's secrets path

### DIFF
--- a/kube-hunter.py
+++ b/kube-hunter.py
@@ -3,6 +3,7 @@ from __future__ import print_function
 
 import argparse
 import logging
+import threading
 
 try:
     raw_input          # Python 2
@@ -66,6 +67,7 @@ def interactive_set_config():
         return False
     return True
 
+
 def parse_docs(hunter, docs):
     """returns tuple of (name, docs)"""
     if not docs:
@@ -74,7 +76,8 @@ def parse_docs(hunter, docs):
     for i, line in enumerate(docs):
         docs[i] = line.strip()
     return docs[0], ' '.join(docs[1:]) if len(docs[1:]) else "<no documentation>"
-    
+
+
 def list_hunters():
     print("\nPassive Hunters:\n----------------")
     for i, (hunter, docs) in enumerate(handler.passive_hunters.items()):
@@ -86,9 +89,14 @@ def list_hunters():
         for i, (hunter, docs) in enumerate(handler.active_hunters.items()):
             name, docs = parse_docs(hunter, docs)
             print("* {}\n  {}\n".format( name, docs))
-        
 
+
+tlock3 = threading.Lock()
+tlock3.acquire()
 hunt_started = False
+tlock3.release()
+
+
 def main():
     global hunt_started 
     scan_options = [
@@ -104,8 +112,10 @@ def main():
 
         if not any(scan_options):
             if not interactive_set_config(): return
-        
+        tlock = threading.Lock()
+        tlock.acquire()
         hunt_started = True
+        tlock.release()
         handler.publish_event(HuntStarted())
         handler.publish_event(HostScanEvent())
         
@@ -117,11 +127,21 @@ def main():
     except EOFError:
         logging.error("\033[0;31mPlease run again with -it\033[0m")
     finally:
+        tlock2 = threading.Lock()
+        tlock2.acquire()
         if hunt_started:
             handler.publish_event(HuntFinished())
             handler.join()
             handler.free()
             logging.debug("Cleaned Queue")
+        tlock2.release()
+
 
 if __name__ == '__main__':
-    main()
+    for i in range(6):
+        try:
+            main()
+        except:
+            import traceback
+            print ('\n\n\n\n\n\n\n\n\n')
+            traceback.print_exc()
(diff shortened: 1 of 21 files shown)


src/modules/hunting/secrets.py
```diff
@@ -0,0 +1,45 @@
...
+
+import requests
+
+from ...core.events import handler
+from ...core.events.types import Vulnerability, Event, OpenPortEvent
+from ...core.types import  Hunter, KubernetesCluster, AccessRisk
+
+
+""" Vulnerabilities """
+class secretsAccess(Vulnerability, Event):
```
lizrice: Should really capitalize class name SecretsAccess for consistency

src/modules/hunting/secrets.py
```diff
@@ -0,0 +1,45 @@
...
+    """ Accessing the pod's secrets within a compromised pod might disclose valuable data to a potential attacker"""
+
+    def __init__(self, evidence):
+        Vulnerability.__init__(self, KubernetesCluster, name="Accessed to pod's secrets", category=AccessRisk)
+        self.evidence = evidence
+
+# Passive Hunter
+#should change the subscribtion here... (openPortEvent isnt relevant..)
```
lizrice: You could do this whenever you find you're running as a pod. Maybe under HostScanEvent, if config.pod is true we should publish a new event called, say, RunningAsPodEvent and use that to trigger this secrets test wdyt?

src/modules/hunting/secrets.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def get_services(self):
+        logging.debug(self.event.host)
+        # get all files and subdirectories files:
+        self.secrets_evidence = [val for sublist in [[os.path.join(i[0], j) for j in i[2]] for i in os.walk('/var/run/secrets/')] for val in sublist]
+        if len(self.secrets_evidence) > 0:
+            return True
+        return False
+
+    def execute(self):
```
lizrice: As commented in PR #54, try to minimize the amount of code that is wrapped in a try...except, otherwise you can miss all sorts of genuine problems. And be very wary of except: pass!

src/modules/hunting/secrets.py
oriagmon: Yea I was going to change it to the HostScanEvent, seems like a great idea to add a RunningAsPodEvent.
Hopefully we won't face that race condition again when subscribing to an event that was thrown from another event..

src/modules/hunting/secrets.py
```diff
@@ -30,16 +30,11 @@ def __init__(self, event):
 
     def get_services(self):
         logging.debug(self.event.host)
+        logging.debug('Passive Hunter is attempting to access pod\'s secrets directory')
         # get all files and subdirectories files:
         self.secrets_evidence = [val for sublist in [[os.path.join(i[0], j) for j in i[2]] for i in os.walk('/var/run/secrets/')] for val in sublist]
-        if len(self.secrets_evidence) > 0:
-            return True
-        return False
+        return True if len(self.secrets_evidence) > 0 else False
```
lizrice: Or `return (len(self.secrets_evidence) > 0)`

kube-hunter.py
```diff
@@ -104,8 +112,10 @@ def main():
 
         if not any(scan_options):
             if not interactive_set_config(): return
-        
+        tlock = threading.Lock()
```
lizrice: I'd prefer we do the lock changes under a separate PR, but since I'm seeing this code here I'll write some thoughts here :-) 

Note that you need just one lock to protect a variable - grab the same lock whenever you're accessing that variable (it needs to be *the exact same lock* wherever you acquire it from). So for a lock that protects a global variable, it also needs to be in global scope.

To make it absolutely obvious what the lock is for, I would 
* create the lock right next to the global variable 
* give it a more meaningful name (like lock_hunt_started or hunt_started_lock)

src/README.md
```diff
@@ -71,7 +71,10 @@ _The file's (module's) content is imported automatically"_
   
 The second step is to determine what events your Hunter will subscribe to, and from where you can get them.  
 `Convention:` Events should be declared in their corresponding module. for example, a KubeDashboardEvent event is declared in the dashboard discovery module.  
-     
+
```
lizrice: This change looks familiar from another PR, so I'd remove it from this one

kube-hunter.py
oriagmon: Alright, I would change it :-) 
But please note I have done the locking code on a [different branch](https://github.com/aquasecurity/kube-hunter/tree/solve-multi-threading-bug) and merged it with that one so I can check it on that hunter (since the threading bug appears here).

src/README.md
oriagmon: Do you mean that one?
"
`Note:` An hunter located under the `disovery` folder should not import any modules located under the `hunting` folder 
in order to prevent circular dependency bugs 
"

src/modules/hunting/secrets.py
```diff
@@ -0,0 +1,40 @@
...
+    def __init__(self, event):
+        self.event = event
+        self.secrets_evidence = ''
+
+    def get_services(self):
+        logging.debug(self.event.host)
+        logging.debug('Passive Hunter is attempting to access pod\'s secrets directory')
+        # get all files and subdirectories files:
+        self.secrets_evidence = [val for sublist in [[os.path.join(i[0], j) for j in i[2]] for i in os.walk('/var/run/secrets/')] for val in sublist]
+        return True if (len(self.secrets_evidence) > 0) else False
```
lizrice: This could just be `return (len(self.secrets_evidence) > 0)`

lizrice: I'm going to merge regardless :)

PR #54, opened 2018-10-09
Service account token hunter

oriagmon:
attempt to access the API server, there might be a room for more hunters on this branch, I will think about it tomorrow ..:)

### DIFF
--- a/src/modules/hunting/apiserver.py
+++ b/src/modules/hunting/apiserver.py
@@ -37,23 +37,29 @@ class AccessApiServerViaServiceAccountToken(Hunter):
 
     def access_api_server(self):
         logging.debug(self.event.host)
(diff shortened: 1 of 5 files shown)


src/modules/hunting/apiserver.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def get_service_account_token(self):
+        logging.debug(self.event.host)
+        with open('/var/run/secrets/kubernetes.io/serviceaccount/token', 'r') as token:
+            data = token.read()
+            self.service_account_token_evidence = data
+        return True
+
+    def execute(self):
+        try:
```
lizrice: It's good practice to put try...except around the smallest amount of code that you think might go wrong. Only use pass VERY sparingly as it could hide all sorts of unexpected things going wrong! So rather than having the try...except here, move it into the tests where things might go wrong. Try to use except to catch the actual things that you expect to go wrong. for example, we should have an "except" case for when we can't read the service account token, which should cause get_service_account_token() to return false

PR #57, opened 2018-10-17
Secrets and multi threading bug

oriagmon:
YAY Multi threading bug is fixed now! :-))))
Tested and succeeded 15 times (previously failed every 4th time ).
Currently, to be 100% sure the bug is indeed solved I am running a bash script that executes the program 100 times so we can discover a threading bug if it exists💃


oriagmon:
UPDATE:
After I have diagnosed the 100 runs I have found 4 of these exceptions  :O 
Seems like the variable locking have helped since it changed the probability of this exception from 0.16-0.25 to 0.04...
I will keep looking on that..

PR #58, opened 2018-10-19
Solve multi threading bug

oriagmon:
I was sure I have already made a PR for that branch on Wednesday, I have even took a pic of it and posted it on my Instagram!
But I'm guessing I did not click the "create pull request" button :O 

**Anyway**, the bug is solved!! It ran successfully 200/200 times.
*The problem was on the self.get() line which was getting and **removing** a queue item, after locking that variable we were able to successfully run the program 200/200 comparing to the 96/100 result we have had before.

* Please have a look on the is_running_lock comments, I wasn't sure if that variable should be locked too..

### DIFF
--- a/kube-hunter.py
+++ b/kube-hunter.py
@@ -99,7 +99,9 @@ hunt_started_lock.release()
 
 
 def main():
-    global hunt_started 
+    hunt_started_lock.acquire()
+    global hunt_started
+    hunt_started_lock.release()
     scan_options = [
         config.pod, 
         config.cidr,
--- a/kube-hunter.py
+++ b/kube-hunter.py
@@ -141,10 +141,5 @@ def main():
 
 
 if __name__ == '__main__':
-    for i in range(1):
-        try:
-            main()
-        except:
-            import traceback
-            print ('\n\n\n\n\n\n\n\n\n')
-            traceback.print_exc()
+        main()
+
--- a/kube-hunter.py
+++ b/kube-hunter.py
@@ -3,6 +3,7 @@ from __future__ import print_function
 
 import argparse
 import logging
+import threading
 
 try:
     raw_input          # Python 2
@@ -87,8 +88,10 @@ def list_hunters():
             name, docs = parse_docs(hunter, docs)
             print("* {}\n  {}\n".format( name, docs))
         
-
+tlock3 = threading.Lock
+tlock3.acquire()
 hunt_started = False
+tlock3.release()
 def main():
     global hunt_started 
     scan_options = [
@@ -104,8 +107,10 @@ def main():
 
         if not any(scan_options):
             if not interactive_set_config(): return
-        
+        tlock = threading.Lock
+        tlock.acquire()
         hunt_started = True
+        tlock.release()
         handler.publish_event(HuntStarted())
         handler.publish_event(HostScanEvent())
         
@@ -117,11 +122,15 @@ def main():
     except EOFError:
         logging.error("\033[0;31mPlease run again with -it\033[0m")
     finally:
+        tlock2 = threading.Lock()
+        tlock2.acquire()
         if hunt_started:
             handler.publish_event(HuntFinished())
             handler.join()
             handler.free()
             logging.debug("Cleaned Queue")
+        tlock2.release()
+
 
 if __name__ == '__main__':
     main()
--- a/kube-hunter.py
+++ b/kube-hunter.py
@@ -129,15 +129,17 @@ def main():
     finally:
         hunt_started_lock.acquire()
         if hunt_started:
+            hunt_started_lock.release()
             handler.publish_event(HuntFinished())
             handler.join()
             handler.free()
             logging.debug("Cleaned Queue")
-        hunt_started_lock.release()
+        else:
+            hunt_started_lock.release()
 
 
 if __name__ == '__main__':
-    for i in range(6):
+    for i in range(1):
         try:
             main()
         except:
--- a/kube-hunter.py
+++ b/kube-hunter.py
@@ -93,15 +93,11 @@ def list_hunters():
 
 global hunt_started_lock
 hunt_started_lock = threading.Lock()
-hunt_started_lock.acquire()
 hunt_started = False
-hunt_started_lock.release()
 
 
 def main():
-    hunt_started_lock.acquire()
     global hunt_started
-    hunt_started_lock.release()
     scan_options = [
         config.pod, 
         config.cidr,
--- a/kube-hunter.py
+++ b/kube-hunter.py
@@ -91,7 +91,7 @@ def list_hunters():
             print("* {}\n  {}\n".format( name, docs))
 
 
-tlock3 = threading.Lock
+tlock3 = threading.Lock()
 tlock3.acquire()
 hunt_started = False
 tlock3.release()
@@ -112,7 +112,7 @@ def main():
 
         if not any(scan_options):
             if not interactive_set_config(): return
-        tlock = threading.Lock
+        tlock = threading.Lock()
         tlock.acquire()
         hunt_started = True
         tlock.release()
@@ -138,4 +138,10 @@ def main():
 
 
 if __name__ == '__main__':
-    main()
+    for i in range(6):
+        try:
+            main()
+        except:
+            import traceback
+            print ('\n\n\n\n\n\n\n\n\n')
+            traceback.print_exc()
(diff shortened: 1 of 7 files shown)


src/modules/report/collector.py
```diff
@@ -3,9 +3,20 @@
...
+services_lock = threading.Lock()
+services_lock.acquire()
 services = list()
+services_lock.release()
+
+global vulnerabilities_lock
+vulnerabilities_lock = threading.Lock()
+vulnerabilities_lock.acquire()
 vulnerabilities = list()
+vulnerabilities_lock.release()
```
danielsagi: Why do that? There is no multithreading when loading of the module.

src/core/events/types/common.py
```diff
@@ -60,28 +66,42 @@ def explain(self):
...
 event_id_count = 0
+event_id_count_lock.release()
+
 """ Discovery/Hunting Events """
+
+
 class NewHostEvent(Event):
     def __init__(self, host, cloud=None):
+        event_id_count_lock.acquire()
         global event_id_count
```
lizrice: As elsewhere, I don't think you need to grab the lock just to declare the variable. Test to make sure I'm correct though!

kube-hunter.py
```diff
@@ -74,23 +76,32 @@ def parse_docs(hunter, docs):
...
             print("* {}\n  {}\n".format( name, docs))
-        
 
+
+global hunt_started_lock
+hunt_started_lock = threading.Lock()
+hunt_started_lock.acquire()
 hunt_started = False
+hunt_started_lock.release()
+
```
lizrice: I think this code will only get called once (when the module is loaded) so you don't really need to grab the lock around it

src/modules/report/collector.py
lizrice: Just to clarify, I think @danielsagi means you don't need to grab & release the lock here, because the code is only called once as the module is loaded. I think it does make sense to have these locks though.

src/modules/report/yaml.py
```diff
@@ -2,41 +2,52 @@
 
 from ruamel.yaml import YAML
 
-from collector import services, vulnerabilities
+from collector import services, vulnerabilities, services_lock, vulnerabilities_lock
+import threading
```
lizrice: do you need to import threading here?

src/modules/report/plain.py
```diff
@@ -3,7 +3,8 @@
 from prettytable import ALL, PrettyTable
 
 from __main__ import config
-from collector import services, vulnerabilities
+from collector import services, vulnerabilities,services_lock, vulnerabilities_lock
+import threading
```
lizrice: do you need to import threading here?

kube-hunter.py
```diff
@@ -74,23 +76,32 @@ def parse_docs(hunter, docs):
...
+
 def main():
-    global hunt_started 
+    hunt_started_lock.acquire()
```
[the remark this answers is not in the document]
oriagmon: Yea, I wasn't sure about it :-) 
I would test it again without these locking policy

src/core/events/handler.py
```diff
@@ -11,8 +11,12 @@
...
 from ...core.events.types import HuntFinished
 
-working_count = 0
-lock = Lock()
+global queue_lock
+queue_lock = Lock()
+
+global is_running_lock
+is_running_lock = Lock()
+
```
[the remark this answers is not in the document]
oriagmon: Indeed it is not used currently, but don't you think we should lock the running variable? I've left the commenting and the locks so you can relate weather it should be locked or shouldn't

src/core/events/types/common.py
oriagmon: I think you are correct, I've added these on the "Using stricter locking "policy"" commit since I didn't understand why the code still has threading issues (when it "evolved" from 3/4 to 96/100 and I couldn't understand why it is not 100/100).

src/modules/report/collector.py
oriagmon: Thanks for the comment Daniel.
I agree with some of your comment- I should review my own code a bit more before submitting a PR (and I would do it the next times!).
But I do prefer locking more variables even if I'm not 100% sure they should be locked Because two main reasons:
1) The fact I think I have fixed the bug, found these exact 2 lines that it coming from and the program runs 100/100 successfully does not mean it would keep acting well in a few weeks when I would add new algorithms that might trigger another multi threading bug..
2) I do prefer better locking too much and having Liz reviewing my work so I can fix it comparing to leaving my code vulnerable to more threading bugs on the future...  (It is easier for a reviewer to see an overuse of the locking than actually checking allover the code for missing locks). 
:-)

src/core/events/types/common.py
oriagmon: -->
Changed & Tested, you were correct :)

src/modules/report/yaml.py
```diff
@@ -2,41 +2,51 @@
...
 from ruamel.yaml import YAML
 
-from collector import services, vulnerabilities
+from collector import services, vulnerabilities, services_lock, vulnerabilities_lock
 
 
 class YAMLReporter(object):
     def get_report(self):
         yaml = YAML()
+        vulnerabilities_lock.acquire()
```
lizrice: Do you need this lock here? There's no obvious access to the variable it protects. (Adding locks where they aren't necessary increases the likelihood that there will be a deadlock introduced.)

lizrice: Do you need this lock here? There's no obvious access to the variable it protects. (Adding locks where they aren't necessary increases the likelihood that there will be a deadlock introduced.)

src/core/events/types/common.py
lizrice: I don't think you need to grab the lock for setting event_id_count to 0 either, because it's only run once when the module is loaded

src/modules/report/collector.py
lizrice: I still don't think you should have to grab the locks here as the code is only run once when the module is loaded

src/modules/report/plain.py
```diff
@@ -14,15 +14,21 @@ class PlainReporter(object):
     def get_report(self):
         """generates report tables"""
         output = ""
+        services_lock.acquire()
```
lizrice: Given there are two locks here, I would definitely prefer this code to only hold each lock briefly to get the length. Having a one-to-one relationship between acquire() and release() makes it easier to read the code. It also means you don't ever have to hold both locks at a time.

src/core/events/types/common.py
```diff
@@ -60,28 +66,40 @@ def explain(self):
         return self.__doc__
 
 
+global event_id_count_lock
+event_id_count_lock = threading.Lock()
+event_id_count_lock.acquire()
```
lizrice: You don't need to grab the lock here, as this is only run once when the module is loaded

src/modules/report/collector.py
lizrice: You don't need to grab the locks here as the code is only run once when the module is loaded. It doesn't do a lot of harm but it's simpler not to do it (and good practice not to grab a lock unless you really need it)

src/core/events/types/common.py
lizrice: You don't need to grab the lock here, as this is only run once when the module is loaded

src/modules/report/plain.py
lizrice: I'd definitely prefer to make this simpler by just grabbing the locks briefly to get the lengths. That way you can have a one-to-one relationship between acquire() and release() which is easier to read. 

Fortunately, you only ever need to hold one lock at a time (which is in general a very good thing - having to hold more than one lock often ends up creating deadlocks). However, at the moment you have to read the code quite carefully to see that's the case. If you use local variables to store len(services) and len(vulnerabilities) it will be really obvious that you only hold one lock at a time.

src/modules/report/yaml.py
lizrice: You don't need this lock here as you're not using the vulnerabilities variable here. 

(And you will go on to grab the lock inside get_vulnerabilities, which is the right place to do it as it's where you're accessing the variable you're protecting)

lizrice: You don't need this lock here (no access of vulnerabilities variable)

lizrice: You don't need this lock here (no access of vulnerabilities variable)

oriagmon: I don't need, I just need to protect the methods that are being called there:
self.get_services(),
self.get_vulenrabilities().
But I'm doing it from within the methods:-)

oriagmon: I don't need, I just need to protect the methods that are being called there:
self.get_services(),
self.get_vulenrabilities().
But I'm doing it from within the methods:-)

PR #60, opened 2018-10-23
More service account token hunters

oriagmon:
Finished with that PR :-) 
I wonder if you think I should delete the "algorithm in words" part

### DIFF
--- a/src/modules/hunting/apiserver.py
+++ b/src/modules/hunting/apiserver.py
@@ -66,6 +66,14 @@ class ListAllRoles(Vulnerability, Event):
                                category=InformationDisclosure)
         self.evidence = evidence
 
+class ListAllRolesUnderDefaultNamespace(Vulnerability, Event):
+    """ Accessing all of the roles under default namespace within a compromised pod might grant an attacker a valuable information
+    """
+
+    def __init__(self, evidence):
+        Vulnerability.__init__(self, KubernetesCluster, name="Access to the all roles list",
+                               category=InformationDisclosure)
+        self.evidence = evidence
 
 class ListAllClusterRoles(Vulnerability, Event):
     """ Accessing all of the namespaces within a compromised pod might grant an attacker a valuable information
@@ -180,7 +188,9 @@ class DeleteAPod(Vulnerability, Event):
 
 
 class ApiServerPassiveHunterFinished(Event):
-    def __init__(self, all_namespaces_names, service_account_token):
+    def __init__(self, all_namespaces_names, service_account_token, host, port):
+        self.host = host
+        self.port = port
         self.all_namespaces_names = all_namespaces_names
         self.service_account_token = service_account_token
 
@@ -203,9 +213,10 @@ class AccessApiServerViaServiceAccountToken(Hunter):
         self.newly_created_cluster_role_name_evidence = ''
         self.newly_created_role_name_evidence = ''
         self.all_namespaces_names_evidence = list()
-        self.all_roles_names_evidence = ''
-        self.all_cluster_roles_names_evidence = ''
-        self.namespaces_and_their_pod_names = dict()
+        self.all_roles_names_evidence = list()
+        self.roles_names_under_default_namespace_evidence = ''
+        self.all_cluster_roles_names_evidence = list()
+        self.namespaces_and_their_pod_names = list(dict())
 
     def access_api_server(self):
         logging.debug(self.event.host)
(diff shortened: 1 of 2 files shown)


src/modules/hunting/apiserver.py
```diff
@@ -0,0 +1,724 @@
...
+
+                    if self.patch_a_role(namespace, self.created_role_evidence):
+                        self.publish_event(PatchARole('Patched Role Name:  {name}'.format(
+                            name=self.patched_newly_created_role_evidence)))
+
+                    if self.delete_a_role(namespace, self.created_role_evidence):
+                        self.publish_event(DeleteARole('Role Status response: {status}'.format(
+                            status=self.deleted_newly_created_role_evidence)))
+
+            # * Algorithm in words: *
```
oriagmon: Do you think we should delete that part?

src/modules/hunting/apiserver.py
```diff
@@ -0,0 +1,716 @@
...
+        self.all_namespaces_names = all_namespaces_names
+        self.service_account_token = service_account_token
+
+    def __str__(self):
+        return str(self.service_account_token)
+
+
+# Passive Hunter
+@handler.subscribe(OpenPortEvent, predicate=lambda x: x.port == 443 or x.port == 6443)
+class AccessApiServerViaServiceAccountToken(Hunter):
```
lizrice: I think I'd have been inclined to split this into 
* a hunter that checks whether you can read the service account token and access the API Server with it (which publishes a ServerApiAccess event) 
* another hunter (or maybe more than one) that subscribe to the ServerApiAccess event 

Just because then each hunter would be a bit easier to read independently

src/modules/hunting/apiserver.py
```diff
@@ -0,0 +1,716 @@
...
+            return res.status_code == 200
+        except (requests.exceptions.ConnectionError, KeyError):
+            return False
+
+    # 1 Namespace method:
+    def get_all_namespaces(self):
+        try:
+            res = requests.get("https://{host}:{port}/api/v1/namespaces".format(host=self.event.host,
+                                                                                port=self.event.port),
+                               headers={'Authorization': 'Bearer ' + self.service_account_token_evidence},
```
lizrice: Since you use the same auth header all over the place, consider storing it in, say, `self.headers` and then you can just have `headers=self.headers` where it's needed

src/modules/hunting/apiserver.py
```diff
@@ -0,0 +1,716 @@
...
+    # 2 Pods Methods:
+    def get_pods_list_under_default_namespace(self):
+        try:
+            res = requests.get("https://{host}:{port}/api/v1/namespaces/default/pods".format(host=self.event.host,
+                                port=self.event.port),
+                               headers={'Authorization': 'Bearer ' + self.service_account_token_evidence}, verify=False)
+
+            parsed_response_content = json.loads(res.content.replace('\'', '\"'))
+            for item in parsed_response_content["items"]:
+                name = item["metadata"]["name"].encode('ascii', 'ignore')
```
lizrice: Why do we need this ascii encoding?

src/modules/hunting/apiserver.py
```diff
@@ -0,0 +1,716 @@
...
+        try:
+            with open('/var/run/secrets/kubernetes.io/serviceaccount/token', 'r') as token:
+                data = token.read()
+                self.service_account_token_evidence = data
+                return True
+        except IOError:  # Couldn't read file
+            return False
+
+    # 2 Pods Methods:
+    def get_pods_list_under_default_namespace(self):
```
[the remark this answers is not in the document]
lizrice: In fact there are a lot of cases here where you do something very like this:

```
try: 
    # build a request
    res = <response from request>
    # do something with response
    return res.content if res.status_code == 200 else False
except (requests.exceptions.ConnectionError, KeyError):
    return False
```

Can you make this cleaner without making it harder to read? Have a look at the kubelet hunter code for inspiration. You could have a class that looks after the basic logic, with the part where you build the request and the part where you handle the response defined for subclasses... 

This may be less of an issue if you can get rid of the quote substitution and the ascii encoding, as there will be less repetitious code.

src/modules/hunting/apiserver.py
oriagmon: The rbac.authorization.k8s.io API return the response in unicode.
I have removed the encode from the other requests :-)

PR #69, opened 2018-12-10
Add basic system:anonymous testing

nhibberd:
Verify `system:anonymous` can not access cluster resources


lizrice:
I love that you've added this hunter @nhibberd! Do you want to make the remaining small changes or would you like us to finish it off?

PR #77, opened 2019-01-17
Correct vulnerability type for API Server access

lizrice:
Unauthenticated access to the API Server is OK - this is allowed for
discovery and health checking. RBAC should prevent access to any API
functionality for anonymous users. So we shouldn't report access to
the API as RCE.


lizrice:
Note this also makes attempts to poll the API even if we don't have a service account token (otherwise we're not checking whether anonymous access is possible!)

lizrice:
This needs more work :-)

PR #82, opened 2019-01-18
Should be able to specify remote address inside pod too

lizrice:
Currently if you specify --pod and --remote, the --remote flag gets ignored. I don't see why you can't specify a particular "remote" address (i.e. a single address to target) even if you're running inside a pod. (This will be helpful if you're trying to debug kube-hunter issues from inside a pod, and you don't need the logs to be full of all the possible pod addresses.)


lizrice:
This change doesn't do what I was originally suggesting when I raised the issue i.e. it adds the ability to target remote addresses as well as the subnet identified in the cluster. But I think it's an improvement and addresses #86 as well

PR #85, opened 2019-02-07
add client ceritficate support

flarno11:
added new option to specify a client certificate to authenticate to the apiserver.

all `requests` invocations that are not used from within the pod have been adjusted

### DIFF
--- a/kube-hunter.py
+++ b/kube-hunter.py
@@ -21,6 +21,8 @@ parser.add_argument('--remote', nargs='+', metavar="HOST", default=list(), help=
 parser.add_argument('--active', action="store_true", help="enables active hunting")
 parser.add_argument('--log', type=str, metavar="LOGLEVEL", default='INFO', help="set log level, options are: debug, info, warn, none")
 parser.add_argument('--report', type=str, default='plain', help="set report type, options are: plain, yaml")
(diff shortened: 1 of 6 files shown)


lizrice:
Hi @flarno11, could you give us some background on this? Is this replicating the scenario where a key has been compromised?

flarno11:
hi @lizrice this is to test the apiserver from outside when you can authenticate only with a client certificate

in our setup we don't reach the kubernetes apiserver if we don't present a valid client certificate to the load balancer in front of kubernetes

lizrice:
@danielsagi could you take a look at this and see what you think?

mcherny:
@lizrice Sorry, it seems I missed your question originally... 
I have two thoughts:

1. It's fine to have it on config. It's not discovered info (as what we put on events), but rather global config applicable to all.
1. There are quite a few  places where we use credentials to access API server etc. I think we should have an utility function rapping the request,get and choosing correct one automatically based on either config or run mode (--pod - credentials inside pod) or .. That way will be less chances to have bugs and non-consistent behavior.

danielsagi:
Hi @flarno11. sorry for taking so long about this. Do you have time to fix the conflicts?

danielsagi:
Closing due to inactivity. We would love to add this feature though. well refer to this in the issue.

PR #86, opened 2019-02-18
enable --cidr option when running as pod

flarno11:
this PR allows the use of the --cidr option together with the --pod option

in our setup the discovery fails because we don't have connectivity to the outside world
with this new option we can just manually specify which IP's we want to test

### DIFF
--- a/src/modules/discovery/hosts.py
+++ b/src/modules/discovery/hosts.py
@@ -76,7 +76,15 @@ class FromPodHostDiscovery(Hunter):
     def execute(self):
         # Discover master API server from in-pod environment variable.
 
-        if self.is_azure_pod():
+        if config.cidr:
+            try:
+                ip, sn = config.cidr.split('/')
+            except ValueError as e:
+                logging.error("unable to parse cidr: {0}".format(e))
+                return
+            cloud = HostDiscoveryHelpers.get_cloud(ip)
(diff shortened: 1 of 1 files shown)


lizrice:
Would it work to just run --cidr without the --pod option? My understanding of this is that --pod enables reading the service account token to simulate attacking from within a compromised pod...or is it useful to use that token to attack the remote hosts including in the CIDR? 

(I'm not completely against this change, I just want to ensure it's useful. We were recently having a similar discussion about whether supporting --remote and --pod together makes sense, and currently I don't think it does, but I could be missing something.)

lizrice:
After discussing with @mcherny we realised there is a neater way to do this, by triggering a HostScanEvent - see #82, which should cover both the --remote and the --cidr case.

flarno11:
the combination of both --cidr and --pod is still useful to simulate the attack from within the cluster, but without relying on the automatic subnet detection (which fails in clusters without internet connectivty)
also it is useful when there are multiple subnets in the cluster

danielsagi:
What you're suggesting, to enable both --pod and --cidr is a good idea.
If you like to, create a PR that enables that, after the changes made in #82 .
This PR's current changes doesn't enable that.

danielsagi:
This functionality was added in #145

PR #92, opened 2019-02-20
created reporter for json format

maniish-jaiin:
This PR adds the ability for the reporter to generate the output in JSON format. JSON format can be consumed easily by Datadog and other metrics and logging platform.

### DIFF
--- a/kube-hunter.py
+++ b/kube-hunter.py
@@ -35,7 +35,7 @@ if config.log.lower() != "none":
 
 from src.modules.report.plain import PlainReporter
 from src.modules.report.yaml import YAMLReporter
-from src.modules.report.json_reporter import JSONReporter
+from src.modules.report.json import JSONReporter
 
 if config.report.lower() == "yaml":
     config.reporter = YAMLReporter()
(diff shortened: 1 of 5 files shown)


kube-hunter.py
```diff
@@ -35,9 +35,12 @@
 
 from src.modules.report.plain import PlainReporter
 from src.modules.report.yaml import YAMLReporter
+from src.modules.report.json_reporter import JSONReporter
```
lizrice: Could we call the module file `json` rather than `json_reporter` for consistency with the other two?

src/modules/report/base.py
```diff
@@ -0,0 +1,34 @@
...
+    def get_services(self):
+        services_lock.acquire()
+        services_data = [{"service": service.get_name(),
+                 "location": "{}:{}{}".format(service.host, service.port, service.get_path()),
+                 "description": service.explain()}
+                for service in services]
+        services_lock.release()
+        return services_data
+
+    def get_vulenrabilities(self):
```
lizrice: I know this typo already existed but while you're here, would you mind correcting the spelling in this function's name to `get_vulnerabilities()`?

kube-hunter.py
maniish-jaiin: Done

src/modules/report/base.py
maniish-jaiin: Done

maniish-jaiin:
All good. Thanks for the help and quick response!

PR #95, opened 2019-02-25
support python3

### LINKED ISSUE #41: Support Python 3
I try kube-hunter using a virtualenv in Python 3 : 

```
$ git clone git@github.com:aquasecurity/kube-hunter.git
$ cd kube-hunter
$ python3 -m venv venv
$ . venv/bin/activate
$ pip3 install -r requirements.txt
[...]
Successfully installed PrettyTable-0.7.2 certifi-2018.8.24 chardet-3.0.4 enum34-1.1.6 idna-2.7 netifaces-0.10.7 requests-2.19.1 ruamel.yaml-0.15.64 scapy-2.4.0 urllib3-1.23
$  python ./kube-hunter.py 
Traceback (most recent call last):
  File "./kube-hunter.py", line 34, in <module>
    from src.modules.report.plain import PlainReporter
  File "/home/rock64/Projects/kube-hunter/src/__init__.py", line 1, in <module>
    import core
ImportError: No module named 'core'
```

westonsteimel:
Adding support for python3

Fixes #41 

Signed-off-by: Weston Steimel <weston.steimel@gmail.com>

### DIFF
--- a/src/modules/discovery/__init__.py
+++ b/src/modules/discovery/__init__.py
@@ -4,4 +4,5 @@ import glob
 # dynamically importing all modules in folder
 files = glob.glob(dirname(__file__)+"/*.py")
 for module_name in (basename(f)[:-3] for f in files if isfile(f) and not f.endswith('__init__.py')):
-    exec('from {} import *'.format(module_name))
+    if not module_name.startswith('test_'):
+        exec('from .{} import *'.format(module_name))
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,3 +7,4 @@ PrettyTable
 urllib3
 ruamel.yaml
 requests_mock
+future
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -1,2 +1,2 @@
-import core
-import modules
+from . import core
+from . import modules
--- a/src/core/__init__.py
+++ b/src/core/__init__.py
@@ -1,2 +1,2 @@
-import types
-import events
+from . import types
+from . import events
--- a/src/core/events/__init__.py
+++ b/src/core/events/__init__.py
@@ -1,2 +1,2 @@
-from handler import *
-import types
+from .handler import *
+from . import types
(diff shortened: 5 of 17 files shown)


westonsteimel:
I'm not sure that this is completely correct yet, but I was able to scan a remote cluster using both python2.7 and python3.7, and I tried json, plain, and yaml as report output formats, and everything appeared to work as expected.

D3n0Duz:
Changes for the Dockerfile
```
FROM python:3.7.2-alpine3.8
```

westonsteimel:
Yes, I wasn't sure if they would want to change that yet or not.  alpine3.9 is available now as well.

westonsteimel:
Also, if preferred I can change everything to use absolute path rather than relative path, I just didn't want to change too much from the way it was originally.  I see for #96 it was changed to an absolute path (and that I'll need to rebase this at some point.)  Thanks!

src/modules/discovery/__init__.py
```diff
@@ -4,4 +4,5 @@
 # dynamically importing all modules in folder
 files = glob.glob(dirname(__file__)+"/*.py")
 for module_name in (basename(f)[:-3] for f in files if isfile(f) and not f.endswith('__init__.py')):
-    exec('from {} import *'.format(module_name))
\ No newline at end of file
+    if not module_name.startswith('test_'):
```
westonsteimel: I suspect this may be unnecessary now due to #96 as well

lizrice: Ha, this might have been a cleverer way of fixing #96!

lizrice:
We'll definitely need to update the Dockerfile eventually as the world moves to Python 3, but I don't mind if that's under this PR or a separate one. 

As to absolute vs relative import paths...I'm not sure which is better (@danielsagi any thoughts?) but I don't think it matters for this PR so long as it works :-)

PR #97, opened 2019-03-03
Cve hunter

idanr1986:
added support for CVE-2019-1002100


src/modules/hunting/cvehunter.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def execute(self):
+        self.get_service_account_token()  # From within a Pod we may have extra credentials
+        api_version = self.get_api_server_version_end_point()
+
+        if api_version:
+            if self.check_cve_2018_1002105(api_version):
+                self.publish_event(ServerApiVersionEndPointAccessPE(self.api_server_evidence))
+
+            elif self.check_cve_2019_1002100(api_version):
```
mcherny:
```suggestion
            if self.check_cve_2019_1002100(api_version):
```

src/modules/hunting/cvehunter.py
```diff
@@ -0,0 +1,121 @@
...
+            version = resDict["gitVersion"].split('.')
+            first_two_minor_digits = eval(version[1])
+            last_two_minor_digits = eval(version[2])
+            return [first_two_minor_digits, last_two_minor_digits]
+
+        except (requests.exceptions.ConnectionError, KeyError):
+            return None
+
+    def check_cve_2018_1002105(self, api_version):
+        first_two_minor_digists = api_version[0]
```
lizrice: In an ideal world we'd correct the spelling in these to all be "digits" rather than "digists", but that's just superficial

PR #101, opened 2019-03-04
API server hunt improvements

lizrice:
Mostly this simplifies a lot of code, while fixing it so that we test both for anonymous access and service account token access. 

We will now get separate events for accessing the API Server anonymously, and accessing it with a service account token. Each of these goes through the passive and then (optionally) active hunter trying to do things through the API. 

Removes some unnecessary API requests that were subsets of others. 

Also adds some more unit tests


src/modules/discovery/apiserver.py
```diff
@@ -11,6 +11,11 @@ class ApiServer(Service, Event):
     def __init__(self):
         Service.__init__(self, name="API Server")
 
+class ApiServerWithServiceAccountToken(Service, Event):
```
mcherny: I guess this event class should be deleted.

PR #105, opened 2019-03-07
Fix #98 - cvehunter now using service token discovered in hosts.py

### LINKED ISSUE #98: Use token we previously obtained
in hunting/cvehunter.py we get the service account token again, but we should be able to reuse the one we obtained in discover/hosts.py

mcherny:
We use the token if available.

### DIFF
--- a/src/modules/hunting/cvehunter.py
+++ b/src/modules/hunting/cvehunter.py
@@ -41,7 +41,6 @@ class IsVulnerableToCVEAttack(Hunter):
         if self.event.auth_token:
             self.headers = {'Authorization': 'Bearer ' + self.event.auth_token}
         self.path = "https://{}:{}".format(self.event.host, self.event.port)
-        self.service_account_token_evidence = ''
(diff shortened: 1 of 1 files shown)


src/modules/hunting/cvehunter.py
```diff
@@ -37,26 +37,20 @@ class IsVulnerableToCVEAttack(Hunter):
     def __init__(self, event):
         self.event = event
         self.headers = dict()
+        # From within a Pod we may have extra credentials
+        if self.event.auth_token:
+            self.headers = {'Authorization': 'Bearer ' + self.event.auth_token}
         self.path = "https://{}:{}".format(self.event.host, self.event.port)
         self.service_account_token_evidence = ''
```
lizrice: I think we no longer use this self.service_account_token_evidence now and could get rid of it

mcherny: @lizrice  You are right. Removed that line. Please merge.

PR #114, opened 2019-04-16
Update imports and dependencies

nshauli:
When using Pyinstaller to create single application kube-hunter, it fails on conflicts between dependencies and missing imports which this pull request is fixing.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,6 @@
 netaddr
 netifaces
-enum34
-scapy
+scapy==2.4.3rc1
 requests
 PrettyTable
 urllib3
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,7 +1,6 @@
 netaddr
 netifaces
-enum34
-scapy
+scapy==2.4.3rc1
```
jerbia: Why are you using a release candidate and not the latest Scapy?

nshauli: There are scapy failures when building kube-hunter image with the latest scapy. These failures were fixed in the release candidate.

PR #116, opened 2019-05-11
Add support for hunters list as part of the reports.

nshauli:
Each reported hunter includes name, description and number of events.
Add severity field to each vulnerability report.
Also changed requirements.txt due to the following error
/home/aqua/.local/lib/python2.7/site-packages/requests/init.py:91: RequestsDependencyWarning: urllib3 (1.25.2) or chardet (3.0.4) doesn't match a supported version!

### DIFF
--- a/src/modules/hunting/cvehunter.py
+++ b/src/modules/hunting/cvehunter.py
@@ -32,7 +32,9 @@ class ServerApiVersionEndPointAccessDos(Vulnerability, Event):
 # Passive Hunter
 @handler.subscribe(ApiServer)
 class IsVulnerableToCVEAttack(Hunter):
-    """ Node is running a Kubernetes version vulnerable to critical CVE-2018-1002105 """
+    """CVE-2018-1002105 hunter
+    Checks if Node is running a Kubernetes version vulnerable to critical CVE-2018-1002105
+    """
 
     def __init__(self, event):
         self.event = event
--- a/src/modules/report/base.py
+++ b/src/modules/report/base.py
@@ -1,4 +1,5 @@
-from .collector import services, vulnerabilities, services_lock, vulnerabilities_lock
+from .collector import services, vulnerabilities, hunters, services_lock, vulnerabilities_lock
+from src.core.types import Discovery
 
 class BaseReporter(object):
     def get_nodes(self):
@@ -22,13 +23,42 @@ class BaseReporter(object):
         services_lock.release()
         return services_data
 
+    def get_severity(self, category):
+        severity = {
+            "Information Disclosure": "medium",
+            "Denial of Service": "medium",
+            "Remote Code Execution": "high",
+            "Identity Theft": "high",
+            "Privilege Escalation": "high",
+            "Access Risk": "low"
+        }
+        return severity.get(category, "low")
+
     def get_vulnerabilities(self):
         vulnerabilities_lock.acquire()
         vulnerabilities_data = [{"location": vuln.location(),
                  "category": vuln.category.name,
+                 "severity": self.get_severity(vuln.category.name),
                  "vulnerability": vuln.get_name(),
                  "description": vuln.explain(),
                  "evidence": str(vuln.evidence)}
                 for vuln in vulnerabilities]
         vulnerabilities_lock.release()
         return vulnerabilities_data
+
+    def get_hunter_statistics(self):
+        hunters_data = list()
+        for hunter, docs in hunters.items():
+            if not Discovery in hunter.__mro__:
+                name, docs = self.get_docs(hunter, docs)
+                hunters_data.append({"name": name, "description": docs, "events": hunter.publishedEvents})
+        return hunters_data
+
+    def get_docs(self, hunter, docs):
+        """returns tuple of (name, docs)"""
+        if not docs:
+            return hunter.__name__, "<no documentation>" 
+        docs = docs.strip().split('\n')
+        for i, line in enumerate(docs):
+            docs[i] = line.strip()
+        return docs[0], ' '.join(docs[1:]) if len(docs[1:]) else "<no documentation>"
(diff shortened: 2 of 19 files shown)


nuwaida:
approved, all unit tests passed, no issue.

src/modules/hunting/cvehunter.py
```diff
@@ -32,7 +32,9 @@ def __init__(self, evidence):
 # Passive Hunter
 @handler.subscribe(ApiServer)
 class IsVulnerableToCVEAttack(Hunter):
-    """ Node is running a Kubernetes version vulnerable to critical CVE-2018-1002105 """
+    """API server version end point and CVE-2018-1002105 hunter
```
lizrice: I am not sure that a comment this long is going to be readable in the output table. Have you checked this running the regular text output, not just the json output?

When you're changing the output please include a "before" and "after" copy of the output so reviewers can see what's changing.

src/modules/hunting/cvehunter.py
```diff
@@ -32,7 +32,9 @@ def __init__(self, evidence):
...
+    Checks if the API server version end point is accessible or node is running a Kubernetes version which vulnerable to critical CVE-2018-1002105
```
lizrice: This is inaccurate / ambiguous - reads as though the result may be showing simply that the API server end point is accessible. Why did this message need to change? 

(Also, grammar-wise, it should be "which is vulnerable").

src/modules/report/base.py
```diff
@@ -22,13 +23,41 @@ def get_services(self):
         services_lock.release()
         return services_data
 
+    def get_severity(self, category):
+        severity = {
+            "Information Disclosure": "medium",
```
lizrice: Are we sure that all information disclosure is medium? @mcherny wdyt?

src/modules/report/base.py
```diff
@@ -22,13 +23,41 @@ def get_services(self):
...
                  "category": vuln.category.name,
+                 "severity": self.get_severity(vuln.category.name),
                  "vulnerability": vuln.get_name(),
                  "description": vuln.explain(),
                  "evidence": str(vuln.evidence)}
                 for vuln in vulnerabilities]
         vulnerabilities_lock.release()
         return vulnerabilities_data
+
+    def get_hunting_hunters_statistics(self):
```
lizrice: get_hunting_hunters_statistics is a confusing name. If we didn't call the Discovery ones DiscoveryHunters, we wouldn't have the semantic problem of distinguishing "real" hunters from the discovery ones, and you can just call it "get_hunter_statistics".

src/modules/report/yaml.py
```diff
@@ -8,8 +8,9 @@ def get_report(self):
         report = {
             "nodes": self.get_nodes(),
             "services": self.get_services(),
-            "vulnerabilities": self.get_vulnerabilities()
+            "vulnerabilities": self.get_vulnerabilities(),
+            "hunting_hunters_statistics": self.get_hunting_hunters_statistics()
```
lizrice: Please call this field "hunter_statistics"

src/modules/report/plain.py
```diff
@@ -88,3 +95,18 @@ def vulns_table(self):
...
+        hunters_table.max_width = MAX_TABLE_WIDTH
+        hunters_table.sortby = "Name"
+        hunters_table.reversesort = True
+        hunters_table.padding_width = 1
+        hunters_table.header_style = "upper"
+
+        hunter_statistics = self.get_hunting_hunters_statistics()
+        for item in hunter_statistics:
+            hunters_table.add_row([item.get("name"), item.get("description"), item.get("events")])
+        return "\nHunting Hunters Statistics\n{}\n".format(hunters_table)
```
lizrice: "Hunter Statistics" please

src/modules/report/json_reporter.py
```diff
@@ -6,6 +6,7 @@ def get_report(self):
         report = {
             "nodes": self.get_nodes(),
             "services": self.get_services(),
-            "vulnerabilities": self.get_vulnerabilities()
+            "vulnerabilities": self.get_vulnerabilities(),
+            "hunting_hunters_statistics": self.get_hunting_hunters_statistics()
```
lizrice: "hunter_statistics" please

src/modules/report/base.py
mcherny: I think that in general severity is by vulnerability and by category. One of my other comments I sent @nshauli is that I don't think severity mapping should be part of reporter.

src/modules/hunting/cvehunter.py
lizrice: I still prefer the original text for this, what was the reason to change it?

src/modules/report/base.py
lizrice: I agree with Micha, if there is a one-to-one mapping between category and severity then this is not adding any information. 

In fact, I think it's worse than that. Isn't this going to return a "low" severity for critical CVE-2018-1002105 which has a category of Privilege Escalation.

src/modules/report/base.py
```diff
@@ -22,13 +23,41 @@ def get_services(self):
...
                  "category": vuln.category.name,
+                 "severity": self.get_severity(vuln.category.name),
                  "vulnerability": vuln.get_name(),
                  "description": vuln.explain(),
                  "evidence": str(vuln.evidence)}
                 for vuln in vulnerabilities]
         vulnerabilities_lock.release()
         return vulnerabilities_data
+
+    def get_hunter_statistics(self):
```
lizrice: Does this duplicate functionality that is already covered [here]?(https://github.com/aquasecurity/kube-hunter/blob/f0160e2eba5f87fd0c5ffdd1681f6eb71dfeaeaf/kube-hunter.py#L99)

Issue #127, opened 2019-05-23
Insecure Azure Cloud IP detection


DrMurx:
To detect whether the IP is in a known public cloud, kube-hunter performs a query to `http://www.azurespeed.com` using an insecure http connection. Given that kube-hunter is supposed to increase cluster security, leaking out internal IP data seems to counter that purpose.

Requests to such an external service should be secure and optional. #107 already suggests to make the check optional to reduce unnecessary timeouts.

DrMurx:
The project which backs the azurespeed site contains the configuration used for the API call.

Check [PublicIPs.xml](https://github.com/blrchen/AzureSpeed/blob/master/src/AzureSpeed.Web.App/Data/IpRangeFiles/Azure/PublicIPs.xml), maybe it can be embedded into kube-hunter?

lizrice:
That’s interesting. I guess we’d need to keep up-to-date with any changes to that file, but it might be more efficient to embed the data

RaitoBezarius:
@lizrice What about using Tor to query the service? Do we care about someone MITM'ing the connection and reading the data we send to the service? Or do we care about getting anonymity during querying the service?

I could send a PR enabling Tor support for external services queries, using SOCKS5 proxy (assumes Tor is running as a proxy, blabla)

DrMurx:
@RaitoBezarius Using Tor for this particular purpose is probably like taking a sledgehammer to crack a nut. A locally running security scanner should not leak internal data to a 3rd party, regardless of the transport, especially if that 3rd party is not an officially supported tool of a known vendor.

RaitoBezarius:
@DrMurx Well, I agree. I guess the simplest way is to embed the PublicIPs data into the tool directly.

itaysk:
the source of truth is here: https://www.microsoft.com/en-us/download/confirmation.aspx?id=56519
This is a json file provided by MS and is updated *weekly*. 

If we want to eliminate external dependencies, we could fetch this file on Kube-Hunter run and write some code that matches against the listed addresses.

blrchen:
I am the owner of [azurespeed.com](http://www.azurespeed.com), you can try to use the alternative endpoint for this api, for example, [https://azurespeed-api-wus.azurewebsites.net/api/region?ipOrUrl=65.52.170.210](https://azurespeed-api-wus.azurewebsites.net/api/region?ipOrUrl=65.52.170.210), this endpoints use SSL cert provided by Azure Web App.

PR #128, opened 2019-05-23
Access cloud IP detection service via HTTPS

DrMurx:
Some mitigation for #127 

azurespeed.com doesn't provide a properly secured API, so this PR changes request to HTTPS without certificate verification.


DrMurx:
It's still far from optimal, also because it might break at any moment. In #127 I made a proposal how to solve this issue.

PR #134, opened 2019-06-05
Added event filtering mechanism

danielsagi:
Added an event filtering mechanism for allowing an interception of specific events, before they are getting published.
All event filters should inherit from `EventFilterBase` class from `src.core.events.types`

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -193,7 +193,9 @@ To prove a vulnerability, create an `ActiveHunter` that is subscribed to the vul
 Sometimes, you may want to filter some events, based on a specific attribute.  
 Creating a filter is simple, the API is similar to creating an Hunter.  
 By inheriting from `EventFilterBase` in `src.core.events.types` module, you can create the filter class.  
-Then, use `@handler.subscribe(Event)` to filter a specific `Event`, kube-hunter's core, will make sure to run your filter on the event before publishing it.   
+Then, use `@handler.subscribe(Event)` to filter a specific `Event`, kube-hunter's core, will make sure to run your filter on the event before publishing it.
+_(You can filter a father event, such as Service or Vulnerability, to filter all services/vulnerabilities)_
+
 #### Similarities to the Hunter API:
 - You need to implement `self.execute(self)` method.
 - to access the event that's being filtered, use `self.event` in your filter.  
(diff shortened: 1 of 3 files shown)


src/core/events/handler.py
```diff
@@ -57,29 +58,66 @@ def subscribe_event(self, event, hook=None, predicate=None):
...
             if hooked_event in event.__class__.__mro__:
                 for hook, predicate in self.hooks[hooked_event]:
                     if predicate and not predicate(event):
                         continue
 
-                    if caller:
-                        event.previous = caller.event
-                        event.hunter = caller.__class__
-
+                    # will run once for each event
```
lizrice: If you only want to run the filter once per event, couldn't the filtering test be done in the loop outside of this loop? Then I don't think you would have to mess with `self.apply_filter.__func__.__code__`

src/README.md
```diff
@@ -189,6 +189,40 @@ The process of proving vulnerabilities, is the base concept of the Active Huntin
...
+    # return True to filter out the event
+    def execute(self):
+        if self.event.category != InformationDisclosure:
+            return True
+```
+The following filter will make sure any vulnerability with a category other than InformationDisclosure will not get published to Kube-Hunters Queue.
+That means other hunters that are subscribed to this vulnerability will not get triggered.
+That opens up a wide variety of possible operations, as this not only can filter out events, but you can actually change event's attributes. and __not__ returning something to indicate you want to keep the changed event.
+
 ## Tests
```
lizrice: Could we have some tests to make sure this works?

src/README.md
```diff
@@ -189,6 +189,40 @@ The process of proving vulnerabilities, is the base concept of the Active Huntin
...
+@handler.subscribe(Vulnerability)
+class InformationDisclosureFilter(EventFilterBase):
+    # return True to filter out the event
+    def execute(self):
+        if self.event.category != InformationDisclosure:
+            return True
+```
+The following filter will make sure any vulnerability with a category other than InformationDisclosure will not get published to Kube-Hunters Queue.
+That means other hunters that are subscribed to this vulnerability will not get triggered.
+That opens up a wide variety of possible operations, as this not only can filter out events, but you can actually change event's attributes. and __not__ returning something to indicate you want to keep the changed event.
```
lizrice: Why would you want to change an event's attributes?

src/README.md
```diff
@@ -189,6 +189,40 @@ The process of proving vulnerabilities, is the base concept of the Active Huntin
...
+- You need to implement `self.execute(self)` method.
+- to access the event that's being filtered, use `self.event` in your filter.  
+
+#### Options for filtering:
+* Remove/Prevent an event from being published
+* Altering event attributes
+
+To prevent an event from being published, return `True` from the execute method of your filter.  
+To alter event attributes, alter the `self.event`, it will change the event itself before it is published.  
+__make sure to not return anything from the execute method if you just want to alter attributes__
```
lizrice: If you're altering attributes should you return None? Or False? Returning False seems more like the opposite of True :-)

danielsagi: You can return False as well, It's basically the same thing, but not returning anything from a function is equal to returning None.

src/core/events/handler.py
danielsagi: @lizrice You're right, that is a simpler solution. but if we run the filter before the loop, in the worst case, where the event is not published, the loop will execute anyways, this way, we make sure the filtering is done only if the event is published, so this is for optimization.

Either case, I'm working on a better solution. will commit it

src/core/events/handler.py
```diff
@@ -70,53 +70,50 @@ def subscribe_event(self, event, hook=None, predicate=None):
...
                         continue
 
-                    # will run once for each event
-                    if self.apply_filter(event):
-                        remove_event = True
-                        break
-                    # formatting the apply_filter method, to not run next time again on the same event.
-                    self.apply_filter.__func__.__code__ = (lambda self, event: None).__code__
-   
+                    # We want to run filters only once and only if we gonna publish it
```
lizrice: Why not keep this simply and just do `apply_filter` on event outside all of this loop, right at the start of this `publish_event` function?

src/README.md
```diff
@@ -202,26 +204,44 @@ Then, use `@handler.subscribe(Event)` to filter a specific `Event`, kube-hunter'
 * Remove/Prevent an event from being published
 * Altering event attributes
 
-To prevent an event from being published, return `True` from the execute method of your filter.  
-To alter event attributes, alter the `self.event`, it will change the event itself before it is published.  
-__make sure to not return anything from the execute method if you just want to alter attributes__  
+To prevent an event from being published, return `None` from the execute method of your filter.  
+To alter event attributes, return a new event, based on the `self.event` after your modifications, it will replace the event itself before it is published.  
+__make sure to return the event from the execute method, or the event will not get publihshed__
```
lizrice: I like these semantics better than the previous version 👍

src/core/events/handler.py
danielsagi: The filtering loop is time consuming, by doing it only when the event gets published, we minimize run time. it's for efficiency purposes.

danielsagi: I see your point. But:
1. If we will have more 'active' hunters the ratio of none-relevant events may change
2. The current implementation (without python internals) is much easier to understand
3. This is 'core' function - that is only few will be changing it (or read for that matter). Little complexity on account of readability is allowed here I think.

So it can go either way. I will leave it to you to decide.

src/core/events/handler.py
```diff
@@ -68,51 +68,42 @@ def subscribe_event(self, event, hook=None, predicate=None):
...
 
-                    logging.debug('Event {} got published with {}'.format(event.__class__, event))
-                    self.put(hook(event))
-            # If the event was filtered out, stop
-            if not event:
-                break
+        # publishing to subscribers
+        if event:
+            # If event was rewritten, make sure it linked to its parent ('previous') event
+            if caller:
```
lizrice: Do we need to do this again, having done it at the start of the function? Maybe it is needed, just checking

danielsagi: Yes, this was suggested by @mcherny , a filter could potentially replace the event entirely, this ensures that no matter what the event chain is kept intact. 
@lizrice

PR #136, opened 2019-06-11
Changed kubelet run handler test to be not a state changing operation

danielsagi:
The way we used to passive check the run handler, has appeared to be state changing.
Looking at the source code of the RunHandler core operation in kubernetes:
https://github.com/kubernetes/kubernetes/blob/8de1569ddae62e8fab559fe6bd210a5d6100a277/pkg/kubelet/kubelet_pods.go#L1584
From there we can see it's enough to pass a non-existent container name, to determine the endpoint is open.

### DIFF
--- a/src/modules/hunting/kubelet.py
+++ b/src/modules/hunting/kubelet.py
@@ -207,14 +207,15 @@ class SecureKubeletPortHunter(Hunter):
 
         # executes one command and returns output
         def test_run_container(self):
+            fake_container_name = "fake-container"
             run_url = self.path + self.Handlers.RUN.value.format(
                 podNamespace=self.pod["namespace"],
                 podID=self.pod["name"],
-                containerName=self.pod["container"],
+                containerName=fake_container_name,
                 cmd = ""
             )
-            status_code = requests.post(run_url, allow_redirects=False, verify=False).status_code 
-            return (status_code != 404 and status_code != 401)
+            output = requests.post(run_url, allow_redirects=False, verify=False).text
+            return output == "container not found (\"{}\")".format(fake_container_name)
 
         # returns list of currently running pods
         def test_running_pods(self):
--- a/src/modules/hunting/kubelet.py
+++ b/src/modules/hunting/kubelet.py
@@ -4,6 +4,7 @@ from enum import Enum
 
 import requests
 import urllib3
+import uuid
 
 from __main__ import config
 from ...core.events import handler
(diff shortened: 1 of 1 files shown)


src/modules/hunting/kubelet.py
```diff
@@ -207,14 +207,15 @@ def test_port_forward(self):
...
             run_url = self.path + self.Handlers.RUN.value.format(
                 podNamespace=self.pod["namespace"],
                 podID=self.pod["name"],
-                containerName=self.pod["container"],
+                containerName=fake_container_name,
                 cmd = ""
             )
-            status_code = requests.post(run_url, allow_redirects=False, verify=False).status_code 
-            return (status_code != 404 and status_code != 401)
+            output = requests.post(run_url, allow_redirects=False, verify=False).text
```
lizrice: I don't think we should be doing *any* POST requests in a passive hunter, so I think this needs to be moved to an active hunter. Wdyt?

danielsagi: basically it's the same as accessing /pods endpoint.
Take a look at the link I posted, because we are making up a container name, we will never pass on to the function that executes something.

lizrice: Agreed it's unlikely that anyone will have a container called "fake_container"... but still, I am not sure this is in the spirit of passive hunters being "read only", as it's attempting to make a state-changing operation, albeit one that it expects to fail. @mcherny wdyt?

danielsagi: I think it's the best check we can do to ensure the run handler is accessible, the alternative is to remove that check and make it only available for the active Hunter

mcherny: @lizrice @danielsagi I tend to agree that even if the POST won't do anything (because the container don't exist) it's still semantically 'write' operation and potentially can cause side effects (e.x. error logs). Hence should not be used by passive hunters.

src/modules/hunting/kubelet.py
```diff
@@ -208,14 +208,13 @@ def test_port_forward(self):
...
+                podNamespace='test',
+                podID='test',
+                containerName='test',
                 cmd = ""
             )
-            status_code = requests.post(run_url, allow_redirects=False, verify=False).status_code
-            # check if return value is 4xx
-            return not 400 <= status_code < 500
+            # if we get a Method Not Allowed, we know we passed Authentication and Authorization.
+            return self.session.get(run_url, verify=False).status_code == 405
```
danielsagi: I did some research and found out if we use a wrong http request method with the right url format for the endpoint, we pass Authentication and Authorization, and get a 405 (method not allowed). so checking that should be enough. wdyt? @lizrice @mcherny 

_Also I have changed the testing to use the self.session, it was a bug, now if running as a pod it uses the token._

PR #140, opened 2019-06-24
Added printing vulnerabilities in case of no services on PlainReporter

danielsagi:
Up until now if services were not discovered, vulnerabilities will not have shown. that made sense, but now we can have vulnerabilities that are not dependent on services. so it's important to have that option enabled.

### DIFF
--- a/src/modules/report/plain.py
+++ b/src/modules/report/plain.py
@@ -26,22 +26,20 @@ class PlainReporter(BaseReporter):
         services_len = len(services)
         services_lock.release()
 
-        if services_len:
-            output += self.nodes_table()
-            if not config.mapping:
-                output += self.services_table()
-                if vulnerabilities_len:
-                    output += self.vulns_table()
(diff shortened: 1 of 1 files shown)


src/modules/report/plain.py
```diff
@@ -40,7 +40,9 @@ def get_report(self):
                     else:
                         output += "\nNo hunters were found"
         else:
-            print("\nKube Hunter couldn't find any clusters")
+            if vulnerabilities_len:
+                output += self.vulns_table()
+            output += "\nKube Hunter couldn't find any clusters"
```
lizrice: I have gone off the idea of saying "couldn't find any Kubernetes services" because in Kubernetes, a service is something different from what we mean here in kube-hunter. Maybe "kube-hunter couldn't find any Kubernetes APIs", wdyt?

danielsagi: I agree,

PR #141, opened 2019-06-24
New kubectl CVE hunter, detecting CVE-2019-11246 and CVE_2019_1002101

danielsagi:
Added: 
* Passive Hunter: "KubectlCVEHunter"
* Discovery Hunter: "KubectlClientDiscovery"
* Component: "KubectlClient"
the hunters, try to get the kubectl version, if kubectl is available in the context of kube-hunter, the PassiveHunter will compare the version to the known fix versions, if it's a vulnerable version, a new vulnerability will rise: appropriate cves vulnerabilities
dependent on #140

### DIFF
--- a/src/core/types.py
+++ b/src/core/types.py
@@ -38,7 +38,7 @@ class KubernetesCluster():
     name = "Kubernetes Cluster"
 
 class KubectlClient():
-    """The kubectl client binary is used by the user, to interact with the cluster"""
+    """The kubectl client binary is used by the user to interact with the cluster"""
     name = "Kubectl Client"
 
 class Kubelet(KubernetesCluster):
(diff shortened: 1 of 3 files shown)


danielsagi:
> This is fine, but shouldn't we also check for versions affected by cve-2019-1002101?

The CVE-2019-11246 is a vulnerability for the incomplete fixes of CVE-2019-1002101. so if you are vulnerable to CVE-2019-11246 you are vulnerable to CVE-2019-1002101, basically the same vulnerability

danielsagi:
> This is fine, but shouldn't we also check for versions affected by cve-2019-1002101?

Youre Right, added changes, now the CVE hunting is more robust

PR #146, opened 2019-07-01
Added new hunter for Capabilities

### LINKED ISSUE #143: CAP_NET_RAW new hunter
Changing kube-hunter to not rely on having a raw socket capability when running in a pod. so not using scapy by default. 
 
Also, adding a hunter which checks for this capability and throws a NET_RAW event should be a good thing

danielsagi:
Added a new hunter "PodCapabilitiesHunter"
Checks for enabled capabilities when running as pod.
For now, publishes a new vulnerability: "CapNetRawEnabled"
 
Closes #143

### DIFF
--- a/src/modules/hunting/capabilities.py
+++ b/src/modules/hunting/capabilities.py
@@ -5,7 +5,7 @@ from ..discovery.hosts import RunningAsPodEvent
 
 from ...core.events import handler
 from ...core.events.types import Event, Vulnerability
-from ...core.types import Discovery, AccessRisk, KubernetesCluster
+from ...core.types import Hunter, AccessRisk, KubernetesCluster
 
 
 class CapNetRawEnabled(Event, Vulnerability):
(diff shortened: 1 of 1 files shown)


src/modules/hunting/capabilities.py
```diff
@@ -0,0 +1,41 @@
...
+
+from ..discovery.hosts import RunningAsPodEvent
+
+from ...core.events import handler
+from ...core.events.types import Event, Vulnerability
+from ...core.types import Hunter, AccessRisk, KubernetesCluster
+
+
+class CapNetRawEnabled(Event, Vulnerability):
+    """CAP_NET_RAW is enabled by default for pods, an attacker could potentially perform network attacks on other pods running on the same node"""
```
lizrice: Better to clarify that this is only a problem if an attacker compromises a pod. How about "CAP_NET_RAW is enabled by default for pods. If an attacker compromises a pod, they can potentially use this capability to perform network attacks on other pods running on the same node"

PR #150, opened 2019-07-07
Added pprof cmdline hunting

### LINKED ISSUE #148: Add pprof hunting on kubelet
Enhance the debugging handlers hunting on the kubelet.

- [x] add pprof/cmdline

danielsagi:
Adds another debug handler hunting, and a new "ExposedKubeletCmdline" vulnerability.
Closes #148


src/modules/hunting/kubelet.py
```diff
@@ -234,6 +242,11 @@ def test_attach_container(self):
             )
             return "/cri/attach/" in self.session.get(attach_url, allow_redirects=False ,verify=False).text
 
+        # returns the cmd line used to run the kubelet
+        def test_pprof_cmdline(self):
+            cmd = self.session.get(self.path + self.Handlers.PPROF_CMDLINE.value)
```
lizrice: In, for example, `test_running_pods` a few lines above we have `verify=False` presumably because we don't know the CAs that the kubelet is using. Do we need it here too? Or perhaps we don't really need it anywhere?

danielsagi: We indeed need it, i've added it now. we should consider setting this in the session variable, so we wont have to add this on each request.

PR #159, opened 2019-08-04
New Hunters: DNS spoofing & ARP spoofing

danielsagi:
Two new hunting modules were added, `arp.py` and `dns.py`
In them are 2 new `Active Hunters` and 2 new `Vulnerabilities`

* __ArpSpoofHunter__ - if a pod has CAP_NET_RAW, checks for the existence of an l3 network traffic on the same node. if there is non, publishing a - `PossibleArpSpoofing`
* __DnsSpoofHunter__ - registers to `PossibleArpSpoofing`, checks if the core-dns(or any dns pod) runs on the same node as normal pods. if so, publishing a - `PossibleDnsSpoofing`


src/modules/hunting/dns.py
```diff
@@ -0,0 +1,54 @@
...
+
+        # getting actuall pod ip of kube-dns service, by comparing the src mac of a dns response and arp scanning.
+        dns_info_res = srp1(Ether() / IP(dst=kubedns_svc_ip) / UDP(dport=53) / DNS(rd=1,qd=DNSQR()), verbose=0)
+        kubedns_pod_mac = dns_info_res.src
+        self_ip = dns_info_res[IP].dst
+
+        arp_responses, _ = srp(Ether(dst="ff:ff:ff:ff:ff:ff")/ARP(op=1, pdst="{}/24".format(self_ip)), timeout=3, verbose=0)
+        for res in arp_responses:
+            if res[1][Ether].src == kubedns_pod_mac:
+                return res[1][ARP].psrc, res[1].src
```
iYehuda: Raise an exception in case of no matches.

src/modules/hunting/dns.py
```diff
@@ -0,0 +1,54 @@
...
+    
+    def execute(self):
+        logging.debug("Attempting to get kube-dns pod ip")
+        self_ip = sr1(IP(dst="google.com", ttl=1), ICMP(), verbose=0)[IP].dst
```
iYehuda: Use a fixed IP address instead of domain name to avoid resolve.

src/modules/hunting/arp.py
```diff
@@ -0,0 +1,52 @@
...
+                    return True
+        # only one mac address for whole LAN and outside 
+        return False
+        
+    def execute(self):
+        self_ip = sr1(IP(dst="google.com", ttl=1), ICMP(), verbose=0)[IP].dst 
+        arp_responses, _ = srp(Ether(dst="ff:ff:ff:ff:ff:ff")/ARP(op=1, pdst="{}/24".format(self_ip)), timeout=3, verbose=0)
+        
+        # arp enabled on cluster and more than one pod on node
+        if len(arp_responses) > 1:
```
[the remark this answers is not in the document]
danielsagi: @iYehuda The check is not for truthfulness, but rather that there is a minimum of 2 arp responses

src/modules/hunting/dns.py
```diff
@@ -0,0 +1,66 @@
...
+        self.event = event
+    
+    def get_cbr0_ip_mac(self):
+        res = srp1(Ether() / IP(dst="1.1.1.1" , ttl=1) / ICMP(), verbose=0)
+        return res[IP].src, res.src
+
+    def extract_nameserver_ip(self):
+        with open('/etc/resolv.conf', 'r') as f:
+            # finds first nameserver in /etc/resolv.conf
+            match = re.search(r"nameserver (\d+.\d+.\d+.\d+)", f.read())
```
iYehuda: `nameserver` may be followed by tabs or more spaces.
Make the pattern to catch any white space after it.

PR #162, opened 2019-08-07
Refactor And Major Bug Fixes in Version and CVE hunting 

### LINKED ISSUE #155: Merge CVE hunters
The following needs to be done for generic cve hunting.
The cvehunter module should contain all hunters which perform version based Vulnerability publishing
- [x] Move kubectl cve hunter to the cvehunter module
- [x] Move generic cve comparison function to utils
- [x] Change api server CVE hunter to use the util function for comparison

danielsagi:
Three major aspects were addressed in this PR:

1. until now, when cve hunting, we performed the check for the version manually. this PR abstracts the mechanism to allow the cve hunting to hunt by registering to `K8sVersionDisclosure`, **that is now a core `Vulnerability` inside `src.core.events.types`. it accepts a `from_endpoint` to allow abstract usage from all hunters that discovers the version of the cluster**

2. All of the CVE hunting was refactored to use one utility function for comparison, and in one 
 new hunting module called `cves.py` the `KubectlCveHunter` was moved there. closes #155 

3. Using this new Utility function, **fixes the bug that crashed the Hunter** when parsing complex versions such as `v1.13.7-gke-rc1`. **now handling all versioning methods**

Also, this PR adds tests for the CVE hunting


src/modules/hunting/proxy.py
```diff
@@ -83,8 +76,8 @@ def execute(self):
         if "buildDate" in version_metadata:
             self.event.evidence = "build date: {}".format(version_metadata["buildDate"])
```
lizrice: Should this be publishing an event too?

src/modules/hunting/apiserver.py
```diff
@@ -557,3 +556,25 @@ def execute(self):
...
+            self.session.headers.update({"Authorization": "Bearer {}".format(self.event.auth_token)})
+        
+    def execute(self):
+        if self.event.auth_token:
+            logging.debug('Passive Hunter is attempting to access the API server version end point using the pod\'s service account token on {}:{} \t'.format(self.event.host, self.event.port))
+        else:
+            logging.debug('Passive Hunter is attempting to access the API server version end point anonymously')
+        version = json.loads(self.session.get(self.path + "/version").text)["gitVersion"]
+        logging.debug("Discovered version of api server {}".format(version))
+        self.publish_event(K8sVersionDisclosure(version=version, from_endpoint="/version"))
```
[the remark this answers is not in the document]
danielsagi: I don't think it's necessary, as the event contains the host and port. the '/version' is the full path from the root anyways

src/modules/hunting/proxy.py
danielsagi: Not really, but perhaps we can join the two, no need to do the same api call twice

PR #163, opened 2019-08-07
New Hunter: Detection for CVE_2019_11247 

danielsagi:
Added a check for CVE_2019_11247. in the `cvehunter.py` module


danielsagi:
Should be rewritten after #162 is merged.

PR #167, opened 2019-08-12
Added Metrics Server Discovery - Distinct from Api Server

danielsagi:
This PR changes the approach we had for discovering API services.

### Motivation
Until now, we confused the Metrics Server with the Api Server, because of their Swagger similarities.

### The Fix
The `ApiServerDiscovery` was changed to `ApiServiceDiscovery`, the main difference is our approach. now this Discovery will publish a new `K8sApiService` Service event.

By utilizing the Filters mechanism, we apply a filter on this event: `ApiServiceClassify`
Which essentially tries to decide what Api service this event exactly is. 
In the case it succeeded, it will replace the event with the appropriate event. 
For now, options are:
* ApiServer
* MetricsServer

If it does not succeed, the event will remain a `K8sApiService`
with the name: "Unrecognized K8s API"


src/modules/discovery/apiserver.py
```diff
@@ -1,41 +1,102 @@
...
+            r = self.session.get("{}://{}:{}".format(protocol, self.event.host, self.event.port))
+            if ('k8s' in r.text) or ('"code"' in r.text and r.status_code is not 200):
+                return True
         except requests.exceptions.SSLError:
             logging.debug("{} protocol not accepted on {}:{}".format(protocol, self.event.host, self.event.port))
         except Exception as e:
             logging.debug("{} on {}:{}".format(e, self.event.host, self.event.port))
 
+
+# Acts as a Filter for services, In the case that we can classify the API,
```
lizrice: I think it might be worth commenting what you are classifying (k8s API server, or metrics server) and how you tell the difference (e.g. we expect k8s API service to respond to a /version end point like this, and the Metrics server to respond like that...)

danielsagi: @lizrice  Done

Issue #164, opened 2019-08-08
Add a subscribe once mechanism to core


danielsagi:
After the PR: #162 will be merged, a potential problem will arise, because the K8sClusterCveHunter could potentially be executed many times, which will cause unnecessary duplicates, **there are many ways to hunt the Api version, but we only want to check that version for CVEs once**

The easiest way, is adding a subscribe once function that acts like the normal subscribe, but will unregister the hunter from the subscribers after it has finished executing.

PR #168, opened 2019-08-12
New Core Feature: Subscribe Once

danielsagi:
Introducing, a subscribe once decorator. 
This allows a hunter to execute only once on the first matched event that's getting published.

### Motivation
When the #162 is merged, the CVE hunter will be triggered on every `K8sVersionDisclosure` event. But in reality we want to only check the version once, we don't care where it came from.

### Implementation and usage
Example:
```python
@handler.subscribe_once(SomeEvent)
class OnceOnlyHunter(Hunter):
    def __init__(self, event):
        self.event = event
    def execute(self):
        print('executed')
```
In this example, no matter how many times `SomeEvent` will get published,
The hunter will run just on the first publish.

The implementation is using the `__new__` magic method in python, to install a `unregister` function on the hook when the decorator function is executed.
so when the first Hunter will be initialized, (actually just before) it will unregister itself from the event

_Tests were added_


danielsagi:
closes #164

Issue #35, opened 2018-08-27
Final hook is hanging


rtoma:
Hi,

Thanks for the tool!

When running aquasec/kube-hunter with `--pod` and `--log=info` inside of a GKE cluster  I get this output:

```
...
~ Started
~ Discovering Open Kubernetes Services...
Event <class 'src.modules.discovery.hosts.HostScanEvent'> got published with <src.modules.discovery.hosts.HostScanEvent object at 0x7fc91436ced0>
Starting new HTTP connection (1): 169.254.169.254:80
http://169.254.169.254:80 "GET /metadata/instance?api-version=2017-08-01 HTTP/1.1" 403 56
Starting new HTTP connection (1): canhazip.com:80
http://canhazip.com:80 "GET / HTTP/1.1" 200 14
Starting new HTTP connection (1): www.azurespeed.com:80
http://www.azurespeed.com:80 "GET /api/region?ipOrUrl=x.x.x.x%0A HTTP/1.1" 200 None
Cannot read wireshark manuf database
```

It eventually stops with `Kube Hunter couldn't find any clusters` (\o/ us), but I am wondering what if the manuf db was available for scapy (?). Maybe something to look into.

lizrice:
This is a benign warning issued by scapy and doesn't affect the behaviour of kube-hunter, but it would be better if we could find a way for it to be avoided.

Joeyn414:
I dont get that much output I'm currently only seeing this:
`~ Started
~ Discovering Open Kubernetes Services...
Cannot read wireshark manuf database`

I did not modify the job.yaml file at all when running it as a job in kubernetes. It just sits there and doesnt do anything else. You state its a benign error but it doesnt move on past this. What can I do to fix whatever is wrong with my setup?

JPLachance:
Hello,

I just tried to deploy the latest version of the kube-hunter job in an AWS EKS cluster and I'm seeing the same behavior described here. In the pod logs, I see:
```
~ Started
~ Discovering Open Kubernetes Services...
Cannot read wireshark manuf database
```
Then nothing else happens.

I tried to deploy the same job in a Kubernetes cluster deployed by KOPS and it worked perfectly.

Maybe GKE and EKS are blocking kube-hunter?

lizrice:
Thank you @JPLachance, we will check that out, it's a good thought. My immediate thought is wondering if it's something to do with security groups that don't allow traffic on these ports. 

[...]

PR #170, opened 2019-08-20
Decrease vulnerabilities on build

danielsagi:
### 2 major changes

1. Removed the wireshark package, scapy was complaining about the manuf file, but in reality this is useless to us, as it just contains mapping of MAC addresses and vendors. Due to the removal of wireshark, the image size decreased significantly. 

2. Changed the base image to be python3.7-alpine3.10 this has the latest updates.
3. added an apk upgrade as well.

#### Minor addition
Because of the removal of wireshark, and the new python version, it was needed to suppress some warnings. so `logging_mod.py` plugin was added. there were added specific scapy and python loggers suppression, mainly to avoid the `Cannot open wireshark manuf...` 

related #35


plugins/logging_mod.py
```diff
@@ -0,0 +1,9 @@
+import logging
+
+# Supress logging from scapy
+logging.getLogger("scapy.runtime").setLevel(logging.CRITICAL)
+logging.getLogger("scapy.loading").setLevel(logging.CRITICAL)
+
+# supress general python warnings
```
lizrice: You're going to remove this right?

Dockerfile
```diff
@@ -1,8 +1,7 @@
-FROM python:3.7.3-alpine3.9 as builder
+FROM python:3.8-rc-alpine3.9 as builder
```
lizrice: just reminding you that maybe we could use Alpine 3.10

PR #206, opened 2019-10-15
document DNS spoofing vulnerability

itaysk:
note that even though the hunter exploits the ARP spoofing vulnerability https://github.com/aquasecurity/kube-hunter/blob/master/docs/kb/KHV020.md, it is still distinct as there could be other ways to spoof. I tried not to mix the two.


itaysk:
added `site.baseurl` as discussed in https://github.com/aquasecurity/kube-hunter/pull/210#discussion_r335545390

Issue #107, opened 2019-03-18
Azure Cloud Checks - Make Optional


goffinf:
I note that there is a check whether the cluster is deployed into Azure Cloud .. see below. 

I don't have a problem with that per se except that in a corporate environment I don't want to request a proxy whitelist exception with my CISO for an endpoint that we have no use for (http://www.azurespeed.com). Moreover, whilst this test will eventually timeout and the checks resume, it delays the process by whatever our default timeout is (120 secs I think) which is not ideal.

Could you make this check optional using something like this ...

```
parser.add_argument('--azurechk', action="store_false", help="whether to check if the cluster is deployed on azure cloud - defaults to false")
...
class HostDiscoveryHelpers:
    @staticmethod
    def get_cloud(host):
        if config.azurechk:
            ...
```

Kind Regards

Fraser.

```
class HostDiscoveryHelpers:
    @staticmethod
    def get_cloud(host):
        if config.azurechk:
            try:
                logging.debug("Checking whether the cluster is deployed on azure's cloud")
                metadata = requests.get("http://www.azurespeed.com/api/region?ipOrUrl={ip}".format(ip=host)).text
            except requests.ConnectionError as e:
                logging.info("- unable to check cloud: {0}".format(e))
                return
        if "cloud" in metadata:
            return json.loads(metadata)["cloud"]
```

shubb30:
I would like to see this too.  I can't have a security tool making calls to the outside.

PR #208, opened 2019-10-15
Feat: Add check to enable azure cloud check

shakeabi:
Fixes: issue #107 .

Review required :)

### DIFF
--- a/kube-hunter.py
+++ b/kube-hunter.py
@@ -18,6 +18,7 @@ parser.add_argument('--log', type=str, metavar="LOGLEVEL", default='INFO', help=
 parser.add_argument('--report', type=str, default='plain', help="set report type, options are: plain, yaml, json")
 parser.add_argument('--dispatch', type=str, default='stdout', help="where to send the report to, options are: stdout, http (set KUBEHUNTER_HTTP_DISPATCH_URL and KUBEHUNTER_HTTP_DISPATCH_METHOD environment variables to configure)")
 parser.add_argument('--statistics', action="store_true", help="set hunting statistics")
+parser.add_argument('--azurechk', action="store_true", help="whether to check if the cluster is deployed on azure cloud - defaults to true")
 
 import plugins
 
--- a/kube-hunter.py
+++ b/kube-hunter.py
@@ -18,7 +18,7 @@ parser.add_argument('--log', type=str, metavar="LOGLEVEL", default='INFO', help=
 parser.add_argument('--report', type=str, default='plain', help="set report type, options are: plain, yaml, json")
 parser.add_argument('--dispatch', type=str, default='stdout', help="where to send the report to, options are: stdout, http (set KUBEHUNTER_HTTP_DISPATCH_URL and KUBEHUNTER_HTTP_DISPATCH_METHOD environment variables to configure)")
 parser.add_argument('--statistics', action="store_true", help="set hunting statistics")
-parser.add_argument('--azurechk', action="store_true", help="whether to check if the cluster is deployed on azure cloud - defaults to true")
+parser.add_argument('--infrastructure-check', action="store_true", help="whether to check if the cluster is deployed on azure cloud - defaults to true")
 
 import plugins
 
--- a/runtest.py
+++ b/runtest.py
@@ -14,6 +14,7 @@ parser.add_argument('--active', action="store_true", help="enables active huntin
 parser.add_argument('--log', type=str, metavar="LOGLEVEL", default='INFO', help="set log level, options are: debug, info, warn, none")
 parser.add_argument('--report', type=str, default='plain', help="set report type, options are: plain, yaml")
 parser.add_argument('--statistics', action="store_true", help="set hunting statistics")
+parser.add_argument('--azurechk', action="store_true", default=True, help="set whether to check if the cluster is deployed on azure cloud - defaults to true | options: True, False")
 
 config = parser.parse_args()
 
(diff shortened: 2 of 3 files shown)


iYehuda:
Thanks for your contribution!

I think it is more reasonable to make this option general for host discovery (i.e. `--discover-cloud`).
Currently there's only check for Azure, but we plan to extend it to other cloud providers.

kube-hunter.py
```diff
@@ -18,6 +18,7 @@
 parser.add_argument('--report', type=str, default='plain', help="set report type, options are: plain, yaml, json")
 parser.add_argument('--dispatch', type=str, default='stdout', help="where to send the report to, options are: stdout, http (set KUBEHUNTER_HTTP_DISPATCH_URL and KUBEHUNTER_HTTP_DISPATCH_METHOD environment variables to configure)")
 parser.add_argument('--statistics', action="store_true", help="set hunting statistics")
+parser.add_argument('--azurechk', action="store_true", help="whether to check if the cluster is deployed on azure cloud - defaults to true")
```
iYehuda: Make this flag more general (i.e. `--infrastructure-check`)

runtest.py
```diff
@@ -14,6 +14,7 @@
 parser.add_argument('--log', type=str, metavar="LOGLEVEL", default='INFO', help="set log level, options are: debug, info, warn, none")
 parser.add_argument('--report', type=str, default='plain', help="set report type, options are: plain, yaml")
 parser.add_argument('--statistics', action="store_true", help="set hunting statistics")
+parser.add_argument('--azurechk', action="store_true", default=True, help="set whether to check if the cluster is deployed on azure cloud - defaults to true | options: True, False")
```
iYehuda: Make this flag more general (i.e. `--infrastructure-check`)

src/modules/discovery/hosts.py
```diff
@@ -54,17 +54,18 @@ def __init__(self, pod=False, active=False, predefined_hosts=list()):
...
-            # azurespeed.com provide their API via HTTP only; the service can be queried with 
-            # HTTPS, but doesn't show a proper certificate. Since no encryption is worse then
-            # any encryption, we go with the verify=false option for the time being. At least
-            # this prevents leaking internal IP addresses to passive eavesdropping.
-            # TODO: find a more secure service to detect cloud IPs
-            metadata = requests.get("https://www.azurespeed.com/api/region?ipOrUrl={ip}".format(ip=host), verify=False).text
-        except requests.ConnectionError as e:
-            logging.info("- unable to check cloud: {0}".format(e))
-            return
+        if config.azurechk:
```
iYehuda: Make this flag more general (i.e. `infrastructure_check`)

kube-hunter.py
```diff
@@ -18,6 +18,7 @@
 parser.add_argument('--report', type=str, default='plain', help="set report type, options are: plain, yaml, json")
 parser.add_argument('--dispatch', type=str, default='stdout', help="where to send the report to, options are: stdout, http (set KUBEHUNTER_HTTP_DISPATCH_URL and KUBEHUNTER_HTTP_DISPATCH_METHOD environment variables to configure)")
 parser.add_argument('--statistics', action="store_true", help="set hunting statistics")
+parser.add_argument('--infrastructure-check', action="store_true", help="whether to check if the cluster is deployed on azure cloud - defaults to true")
```
iYehuda: I think it would be better to disable this functionality with this flag, e.g.:
```python
parser.add_argument('--no-infrastructure-check', action="store_false", ...)
```

runtest.py
```diff
@@ -14,6 +14,7 @@
 parser.add_argument('--log', type=str, metavar="LOGLEVEL", default='INFO', help="set log level, options are: debug, info, warn, none")
 parser.add_argument('--report', type=str, default='plain', help="set report type, options are: plain, yaml")
 parser.add_argument('--statistics', action="store_true", help="set hunting statistics")
+parser.add_argument('--infrastructure-check', action="store_true", default=True, help="set whether to check if the cluster is deployed on azure cloud - defaults to true | options: True, False")
```
iYehuda: Remove `default=True`.
Using it with `store_true` action basically does nothing, it will be true whether the flag is set or not.
Nevertheless, this flag should be disabling the check (as commented before)

src/modules/discovery/hosts.py
```diff
@@ -54,17 +54,18 @@ def __init__(self, pod=False, active=False, predefined_hosts=list()):
...
-            # azurespeed.com provide their API via HTTP only; the service can be queried with 
-            # HTTPS, but doesn't show a proper certificate. Since no encryption is worse then
-            # any encryption, we go with the verify=false option for the time being. At least
-            # this prevents leaking internal IP addresses to passive eavesdropping.
-            # TODO: find a more secure service to detect cloud IPs
-            metadata = requests.get("https://www.azurespeed.com/api/region?ipOrUrl={ip}".format(ip=host), verify=False).text
-        except requests.ConnectionError as e:
-            logging.info("- unable to check cloud: {0}".format(e))
-            return
+        if config.infrastructure_check:
```
iYehuda: Use negation with the new flag (`no_infrastructure_check`)

itaysk:
@iYehuda regarding your suggestion to make this about cloud not about azure, wouldn't it make sense to make it about external access?
Iooking at related issues, I don't think people mind that kube-hunter is discovering they are running in a cloud, they care that we are using an external service to discover the public IP. Adding a "discover-cloud" or "infrastructure-check" are not fixing the concern because an unrelated hunter could use an external service for something else.
Wouldn't it be better to gate this feature with a "allow-external-dependencies" or something the captures the concern and can be useful across the board?

itaysk:
Thanks for working on this, I'm afraid we are going to close this PR in favor of another PR that refactors the entire IP detection, that would be the better place to fix this https://github.com/aquasecurity/kube-hunter/pull/270 . Also given my previous comment https://github.com/aquasecurity/kube-hunter/pull/208#issuecomment-552859128 we will probably even fix this in a broader scope in another PR. Thanks again

PR #210, opened 2019-10-16
fix navigation url when searching for kb article

itaysk:
when hosted under a path (for example, `http://host/sub`), the composed link doesn't account for the path (`/sub`). This PR fixes this by using Jekyll's `link` tag to build the URL

### DIFF
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -62,7 +62,7 @@
     </div>
     <script src="{{ "/assets/js/scale.fix.js" | relative_url }}"></script>
     <script type="text/javascript">
-        var articleUrlPrefix = "{{ site.url }}/kb/";
+        var articleUrlTemplate = "{% link kb/KHV002.md %}";
         var searchInput = document.getElementById("searchInput");
         var searchButton = document.getElementById("searchButton");
         searchInput.addEventListener("keyup", function(event) {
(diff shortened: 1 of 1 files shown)


docs/_layouts/default.html
```diff
@@ -62,7 +62,7 @@ <h1><a href="{{ "/" | absolute_url }}">
     </div>
     <script src="{{ "/assets/js/scale.fix.js" | relative_url }}"></script>
     <script type="text/javascript">
-        var articleUrlPrefix = "{{ site.url }}/kb/";
+        var articleUrlTemplate = "{% link kb/KHV002.md %}";
```
iYehuda: `link` does not include site base url.
You should prepend it to the link:
`var articleUrlTemplate = "{{ site.baseurl }}{% link kb/KHV002.md %}";`

itaysk: This was a confusing one. When I tested this locally it worked fine. Also when I read the documentation on the `link` tag, it said that adding the base url is not required since Jekyll `4.0.0`: https://github.com/jekyll/jekyll/blame/master/docs/_docs/liquid/tags.md#L88 but then I run a test on another GitHub Pages and saw that the generated link was missing the base url.
[Turns out](https://pages.github.com/versions/) that the Jekyll version that GitHub uses now is `3.8.5` which explains why this is required. 
Thanks for catching this!

iYehuda:
lgtm

Issue #186, opened 2019-10-01
Why is kube-hunter showing a vulnerability in a version that is patched for the CVE?


chrisante:
![image](https://user-images.githubusercontent.com/42646545/65976619-611e6d00-e43e-11e9-90e2-69e69f7b9204.png)

EKS version is the correct patched version yet, the following CVE's (CVE-2019-9514 and CVE-2019-9512) are returned as vulnerabilities. Is there possibly an issue with the Go version/worker nodes that's not up to date with the api server? 

CVE Reference:https://discuss.kubernetes.io/t/security-release-of-kubernetes-v1-15-3-v1-14-6-v1-13-10-cve-2019-9512-and-cve-2019-9514/7596

lizrice:
@iYehuda I think you are already looking at this, right?

PR #220, opened 2019-10-18
Consider patched versions as not vulnerable by default

### LINKED ISSUE #194: Default value for --ignore-downstream flag
If you don't specify the --ignore-downstream flags, you will get false positives for patched versions of Kubernetes. Ideally we would know the patch identifiers so we can report accurately, but since kube-hunter is not intended to be a comprehensive CVE checker, that's not critical. It would be better if the default behaviour were to NOT report on these false positives.

iYehuda:
Change `--ignore-downstream` to `--ignore-patched-versions` and
invert it's effect.
From now on, kube-hunter will not alert patched components as default
behavior.

Resoles #186 
Resolves #194


itaysk:
do you think we need to rename the `ignore_downstream` in the `is_vulnerable` function accordingly?

iYehuda:
> do you think we need to rename the `ignore_downstream` variable in the `is_vulnerable` function accordingly?

I tried to make minimal change to address issue #194 .
Beside that, I think the flag `--include-patched-versions` describes higher level meaning than string comparison as in `is_vulnerable` function.
It is possible that `ignore_downstream` may not be the best fit but this argument is out of this PR scope.

What do you think? @itaysk

itaysk:
ok

PR #223, opened 2019-10-19
Implemented PEP8 over Python files

vipulgupta2048:
Huge improvements in the readability of code by the implementation of best practices over the Kube-Hunter code. No worries if the maintainers think its best to close it. Often changes like these should be discussed before. But, I feel actions speak louder than issues discussions.

Some changes might be a little drastic, and I am willing to change them if the maintainers wish to. Thanks, 

Signed-off-by: Vipul Gupta (@vipulgupta2048) <vipulgupta2048@gmail.com>

### DIFF
--- a/install_imports.py
+++ b/install_imports.py
@@ -1,15 +1,18 @@
 from os.path import basename
 import glob
 
+
 def get_py_files(path):
     for py_file in glob.glob("{}*.py".format(path)):
         if not py_file.endswith("__init__.py"):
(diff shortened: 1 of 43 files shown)


iYehuda:
Hi @vipulgupta2048 !

I agree with your PRs target, we should reorganize the code base to meet standard style guides.
However, reformatting the code base should be done as an ongoing process alongside other changes rather than a single commit.

Thanks for your work!

PR #238, opened 2019-10-22
build article titles from metadata

itaysk:
## Description (*)
build article titles from metadata instead of writing it twice.
also changed `id` in metadata to `vid` to avoid conflict with Jekyll's `page.id`


itaysk:
I got a notification that the Jekyll build failed after this merge. It is working locally for me using this branch. The kb is still up using prev version, I'll investigate.

Issue #240, opened 2019-10-23
Add search for the KB


itaysk:
Currently the search is more like a lookup, where you can lookup an article by it's ID. We need a full text search to allow people to search for content freely.

iYehuda:
Can we run some backend logic on GitHub pages to do that?
What is the common way to do this on that platform?

itaysk:
No. We can do client side, or server side hosted as a service. There are several of those that work with Jekyll like algolia/lunr just to give an example

RobbiNespu:
how about something like this
![Screenshot-20191024102538-962x673](https://user-images.githubusercontent.com/3077875/67448126-a5b9b480-f648-11e9-8c4e-ca1275895daf.png)

itaysk:
looks good, but should replace or integrate with the current search in the sidebar. 
better to discuss the behavior and design of the feature here, and the implementation in the PR, so I'm adding [WIP] to your PR just so maintainers don't review it yet. Please remove it once you are ready to review.

itaysk:
I think it would be best to use some external search indexing service. One option is Google custom search. It's free but ad supported. Another is Algolia which is free for not-commercial project which I guess we qualify.

RaitoBezarius:
That sounds overkill, this seems a more appropriate solution if possible: http://elasticlunr.com/

itaysk:
@RaitoBezarius if that can support full text indexing (of the content) that's an option too

PR #244, opened 2019-10-24
full text search

RobbiNespu:
## Description
Allow to searching full text on all KB pages  as issue #240 

## Contribution Guidelines

## Contribution checklist
 - [ /] I have read the Contributing Guidelines.
 - [ /] The commits refer to an active issue in the repository.
 - [ X] I have added automated testing to cover this case.
 
## Notes


itaysk:
Thanks for the PR! can you please incorporate the results into the existing search box? if it can replace the existing functionality that's fine too.

RobbiNespu:
It much better to replace since it also can search the ID. The best design is to put on the top of content because result are show in list. since sidebar design is fixed. It will be ugly a bit.

RobbiNespu:
image

RobbiNespu:
Hi @itaysk , I commit some changes as mentioned. Could you check if satisfying the expectation?

itaysk:
Thanks @RobbiNespu for your efforts. First, I cannot run this locally, getting some client side errors. But regardless, I now see that this is not in fact a full-text search, because it only search the metadata of articles. It's definitely an improvement over the current lookup by only vid, but I think if merged, this PR will quickly be superseded by another PR that adds full-text search so I don't see value in investing in this one now.
(This is a good lesson to follow the contribution guidelines and first discuss the feature in an Issue before submitting a PR)
I think a better solution would be to leverage an external search service for indexing, so I'm deferring the discussion to the associated issue. Please feel free to discuss the solution over there. Once we agree on a solution you can submit a new PR.

lizrice:
Closing based on @itaysk comment

PR #260, opened 2019-10-29
docs: Vulnerability Types Description

### LINKED ISSUE #253: Document categories
we need to add some documentation of the existing categories and their meaning. 
Also, maybe worth to report the relation to severity until that is changed (https://github.com/aquasecurity/kube-hunter/issues/118)

MiniMarvin:
## Description
Added a description for each vulnerability type that the kube-hunter can find up to now. That's important so each contributor can understand properly which type of vulnerability should use on writing code without the need to look for examples or what that vulnerability type means. 

## Contribution Guidelines

## Fixed Issues

closes #253 

Fixes #(issue)

### BEFORE
Any Terminal Output Before Changes.

### AFTER
Any Terminal Output Before Changes.

## Contribution checklist
 - [X] I have read the Contributing Guidelines.
 - [X] The commits refer to an active issue in the repository.
 
## Notes
As the updates where only in documentation no automation test was written.

### DIFF
--- a/src/core/types.py
+++ b/src/core/types.py
@@ -51,31 +51,75 @@ class Azure(KubernetesCluster):
     name = "Azure"
 
 
-""" Categories """
+""" Categories 
+
+Defines the category of each vulnerability 
+That could be exploited by an attacker.
+"""
 class InformationDisclosure(object):
+    """ Security issue where the attacker can
+    gain access to sensitive information in
+    the Virtual Machine. 
+    """
     name = "Information Disclosure"
 
 
 class RemoteCodeExec(object):
+    """ Remote execution of code, it is, you can
+    execute code in the VM from anywhere, 
+    wether arbitrary code or specific code.
+    """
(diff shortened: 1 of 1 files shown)


iYehuda:
Hi @MiniMarvin !

I think it is more reasonable to make a reference to more standard definitions of vulnerability and weakness types, see [MITRE CWE](https://cwe.mitre.org/data/definitions/2000.html) for example.
Apart from that, your definitions are about VMs which is more an example than definitions.
I wish to give more general, yet precise definition to newcomers.

iYehuda:
Hi @MiniMarvin !

Any progress with this?
I think adding links to external and standard definitions would be a good idea.
What do you think @lizrice @itaysk ?

lizrice:
Agreed, we should use standard definitions for these rather than making up our own

lizrice:
Closing this PR since we would use a different approach

PR #264, opened 2019-10-30
Remove __main__ references and create a top-level config module

RaitoBezarius:
## Description
It creates a top-level configuration module named `conf` which can be used without `__main__` reference, and reused throughout the application to provide the configuration.

It enables external profiling (memory profiling, CPU profiling for example), all scripts which have a different entry point than the kube-hunter script but still call it are now usable.

I didn't replace the `runtests.py` for now because I felt it was too much for a PR, this is simpler to grok right now, and when it feels appropriate, we could try to see how to replace directly the script with an good `pytest` call.

## Fixed Issues
Relates to #234 

## Contribution checklist
 - [x] I have read the Contributing Guidelines.
 - [x] The commits refer to an active issue in the repository.
 
## Notes
I removed the stuff about automated testing because this change has no impact.

PR #270, opened 2019-10-31
Initial design for ExternalService API with service for public IP

### LINKED ISSUE #235: Dependence on canhazip.com for external IPs fetch

Currently, in Discovery Hosts module, canhazip.com is used as an external service, it'd be interesting to add the possibility to replace it by another service (just provide a canhazip.com—compatible service or some modular way to have an "ExternalIPService" instance which can give your external IP address).

## Why is this needed

In the same vein as #127
This is not a good idea to depend on an external service which could log the request made to it or some third-party sitting on the link could listen to this request (assuming DNS requests are clear and can be listened), so for a blackbox pentesting in a company, it could be detected.

RaitoBezarius:
## Description
As per to #235 ; I'm trying to see what kind of external service API would fit the usage in Kube Hunter (as in, it's generalizable for other external services).
Here is an initial idea, I didn't want to go full OO magic, using singletons pattern & whatever.

This is a very simple PR, though it has the potential for external plugins to use their own providers of public IP, it can be used to modify if you want IPv4/IPv6, it can accommodate special use cases: when you want *private* IP, let's say, you're in a private VPC or something like this.

Two things are left as questions and require feedback:

- Test, how do we want to go? I can test for generic failures using mock and ensure that `try_get` calls the second provider. That's a simple enough and relevant test to do. But before, I want to know if that'd be interesting to move this logic into a base class called `ServiceWithFailover` which enables this pattern.
- Propagate a form of configuration to the modules, if we want to globally use only this provider, we may want that all subsequent instances of ExternalService instantiates using only those specific providers, what I usually suggest is to use module name which are importable: `["plugins.external_providers.canhazip", "plugins.external_providers.ipify"]` for example, this can be configured in some file or even in the `conf.py` if #264 is merged.

This enables new patterns to configure stuff using environment variables too. Anyway, I need feedback to understand what maintainers feel is the most appropriate for this project.

Bonus: Ipify.org is added.

## Fixed Issues

Fixes #235 

## Contribution checklist
 
## Notes
As above-mentioned, automated testing is a question.

### DIFF
--- a/src/modules/external_services/exceptions.py
+++ b/src/modules/external_services/exceptions.py
@@ -1,7 +1,15 @@
 class ExternalServiceFailure(Exception):
+    """
+    This exception can be thrown when a service action has failed (e.g. unable to determine any IP address, no cache result available when offline)
+    It is for user-code and external consumption of the service.
+    """
     def __init__(self):
         pass
 
 class ExternalProviderFailure(Exception):
+    """
+    This exception can be thrown when the provider of the service fails (e.g. the site is down, 400 error, etc.)
+    It is for internal consumption of external services which will decide to act upon a provider failure, by providing a cached result or throw a service failure.
+    """
     def __init__(self, exc=None):
         self.wrapped_exc = exc
--- a/src/modules/external_services/exceptions.py
+++ b/src/modules/external_services/exceptions.py
@@ -1,0 +1,7 @@
+class ExternalServiceFailure(Exception):
+    def __init__(self):
+        pass
+
+class ExternalProviderFailure(Exception):
+    def __init__(self, exc=None):
+        self.wrapped_exc = exc
--- a/src/modules/external_services/external_ip.py
+++ b/src/modules/external_services/external_ip.py
@@ -11,37 +11,41 @@ def generic_text_http_provider(url):
     except requests.ConnectionError as e:
         raise ExternalProviderFailure(e)
 
-ExtIPProvider = namedtuple('ExternalIPProvider', [
-    'handler',
+ExternalIPProvider = namedtuple('ExternalIPProvider', [
+    'provide', # provide the current address IP.
     'priority' # it means: secure, no logs, etc.
 ])
 
-CanHazIP = ExtIPProvider(handler=generic_text_http_provider('https://canhazip.com'), priority=0)
-Ipify = ExtIPProvider(handler=generic_text_http_provider('https://api.ipify.org'), priority=1)
+CanHazIP = ExternalIPProvider(provide=generic_text_http_provider('https://canhazip.com'), priority=0)
+Ipify = ExternalIPProvider(provide=generic_text_http_provider('https://api.ipify.org'), priority=1)
 
 DEFAULT_PROVIDERS = [CanHazIP, Ipify]
 
 class ExternalIPService:
     def __init__(self, providers=None, cache_results=True):
-        if providers and not all(isinstance(i, ExtIPProvider) for i in providers):
+        if providers and not all(isinstance(i, ExternalIPProvider) for i in providers):
             raise TypeError('A provider must be of type `ExternalIPProvider`!')
-        
+
         # Order provider by priority
         self.providers = sorted(providers or DEFAULT_PROVIDERS, key=lambda provider: provider.priority, reverse=True)
         self.cache_results = cache_results
+        self.cached_ip = None
 
-    def try_get(self, bypass_cache=False):
-        if not bypass_cache and self.cache_results and self.cached_ip:
+    def try_get(self, use_cache=True):
+        if use_cache and self.cache_results and self.cached_ip:
             return self.cached_ip
-        
+
         last_exc = None
         for provider in self.providers:
             try:
-                ip = provider.handler()
-                self.cached_ip = ip
+                ip = provider.provide()
+
+                if self.cache_results:
+                    self.cached_ip = ip
+
                 return ip
             except ExternalProviderFailure as exc:
                 last_exc = exc
-        
+
         # Tough luck.
         raise ExternalServiceFailure(exc=last_exc)
--- a/src/modules/external_services/external_ip.py
+++ b/src/modules/external_services/external_ip.py
@@ -1,7 +1,9 @@
-from .exceptions import ExternalServiceFailure, ExternalProviderFailure
+from collections import namedtuple
 
 import requests
-from collections import namedtuple
+
+from .exceptions import ExternalProviderFailure, ExternalServiceFailure
+
 
 def generic_text_http_provider(url):
     try:
--- a/src/modules/external_services/external_ip.py
+++ b/src/modules/external_services/external_ip.py
@@ -1,0 +1,45 @@
+from .exceptions import ExternalServiceFailure, ExternalProviderFailure
+
+import requests
+from collections import namedtuple
+
+def generic_text_http_provider(url):
+    try:
+        return requests.get(url).text
+    except requests.ConnectionError as e:
+        raise ExternalProviderFailure(e)
+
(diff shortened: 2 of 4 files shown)


lizrice:
Sorry it has taken so long to get back to you on this @RaitoBezarius - we had a lot of Hacktoberfest changes and then KubeCon. I like the general idea of this, will leave it with @iyehuda and @itaysk to figure out how they want to take this forward. 

I realized we could be querying http**s**://canihazip.com/**s** which returns simply the IP address, which I think would be better than http://canihazip.com. Which then begs the question whether it's worth falling back between one service and another identical one, but I suppose one of them might be down.

RaitoBezarius:
@lizrice No problem, I have some cleanup to do myself on my contributions. Regarding the HTTPS vs HTTP stuff, it's baked in this PR (except the final `/s` I suppose), this was a low hanging fruit I guess which is cool.
One of them might be down, worst, one of them might be blocked (?).

To me, as a red teamer, this is the kind of the tool I'd use for initial recognition (it's super cool and I really appreciate Python was used for this because it makes it easier for me to extend than Ruby/Perl stuff I deal with), I have done some tests with it and it gave really interesting results.

In a corporate environment or in certain super strange contexts, it is really important for me to control that kind of service.
Better, it enables me to switch a public IP service for a private IP service, which is really neat in cloud infrastructures (if it makes sense for the tool or during an assessment).

So… I guess I'll just do a real effort and clean my PRs so you guys can see how do you want to move forward with this.

src/modules/external_services/external_ip.py
```diff
@@ -0,0 +1,45 @@
...
+import requests
+from collections import namedtuple
+
+def generic_text_http_provider(url):
+    try:
+        return requests.get(url).text
+    except requests.ConnectionError as e:
+        raise ExternalProviderFailure(e)
+
+ExtIPProvider = namedtuple('ExternalIPProvider', [
```
iYehuda: Rename to `ExternalIPProvider`

src/modules/external_services/external_ip.py
```diff
@@ -0,0 +1,45 @@
...
+class ExternalIPService:
+    def __init__(self, providers=None, cache_results=True):
+        if providers and not all(isinstance(i, ExtIPProvider) for i in providers):
+            raise TypeError('A provider must be of type `ExternalIPProvider`!')
+        
+        # Order provider by priority
+        self.providers = sorted(providers or DEFAULT_PROVIDERS, key=lambda provider: provider.priority, reverse=True)
+        self.cache_results = cache_results
+
+    def try_get(self, bypass_cache=False):
```
iYehuda: I think it is better to have `use_cache` parameter with default value of `True` since you check if cache is wanted

src/modules/external_services/exceptions.py
```diff
@@ -0,0 +1,7 @@
+class ExternalServiceFailure(Exception):
+    def __init__(self):
+        pass
+
+class ExternalProviderFailure(Exception):
+    def __init__(self, exc=None):
+        self.wrapped_exc = exc
```
iYehuda: The difference between these exceptions is not clear (I understood that after reading the using code).
Please add docstring that will clarify the usage and difference

src/modules/external_services/external_ip.py
```diff
@@ -0,0 +1,45 @@
...
+        if not bypass_cache and self.cache_results and self.cached_ip:
```
iYehuda: `self.cached_ip` should be defined in `__init__`

src/modules/external_services/external_ip.py
```diff
@@ -0,0 +1,45 @@
...
+from collections import namedtuple
+
+def generic_text_http_provider(url):
+    try:
+        return requests.get(url).text
+    except requests.ConnectionError as e:
+        raise ExternalProviderFailure(e)
+
+ExtIPProvider = namedtuple('ExternalIPProvider', [
+    'handler',
```
iYehuda: Use more imperative name (e.g. `get`/`fetch`/`provide`) to make the usage more clear

src/modules/external_services/external_ip.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def try_get(self, bypass_cache=False):
+        if not bypass_cache and self.cache_results and self.cached_ip:
+            return self.cached_ip
+        
+        last_exc = None
+        for provider in self.providers:
+            try:
+                ip = provider.handler()
+                self.cached_ip = ip
```
iYehuda: Cache the result if `self.cache_results` is `True`
It has no major effect with IP address string but it can get more memory intensive with other data services.

src/modules/external_services/external_ip.py
```diff
@@ -0,0 +1,45 @@
...
+    'handler',
+    'priority' # it means: secure, no logs, etc.
+])
+
+CanHazIP = ExtIPProvider(handler=generic_text_http_provider('https://canhazip.com'), priority=0)
+Ipify = ExtIPProvider(handler=generic_text_http_provider('https://api.ipify.org'), priority=1)
+
+DEFAULT_PROVIDERS = [CanHazIP, Ipify]
+
+class ExternalIPService:
```
iYehuda: I think this class can be a general `BaseExternalService` with failover in it so it can be reused

iYehuda:
Hi @RaitoBezarius !

Any progress made?

RaitoBezarius:
> Hi @RaitoBezarius !
> 
> Any progress made?

Hi there, I addressed most of the comments, I'll fix the conflicts and I need to ask you regarding the base class stuff.

src/modules/external_services/external_ip.py
RaitoBezarius: Do you mean like, I should go ahead and generalize it for any type of results?

RaitoBezarius: I felt like I could do it, but I was wondering if it was a really good idea, but if you want I can do the modifications and make it a `external_base_service.py` and derive `external_ip.py` from it.

iYehuda: It feels right to me, I think it will save some code duplication in future.
Let's go with a `base_external_service.py` with `BaseExternalService` class.

src/modules/external_services/external_ip.py
```diff
@@ -0,0 +1,51 @@
...
+    except requests.ConnectionError as e:
+        raise ExternalProviderFailure(e)
+
+ExternalIPProvider = namedtuple('ExternalIPProvider', [
+    'provide', # provide the current address IP.
+    'priority' # it means: secure, no logs, etc.
+])
+
+CanHazIP = ExternalIPProvider(provide=generic_text_http_provider('https://canhazip.com'), priority=0)
+Ipify = ExternalIPProvider(provide=generic_text_http_provider('https://api.ipify.org'), priority=1)
```
iYehuda: I don't think using priority fits here.
In my opinion, the service provider mechanism should be as easy as possible to use.
Maybe just use a list of them so first item is first priority?

RaitoBezarius: Sure, but if some user want to dynamically modify the providers and want to have his own to be used rather than the system ones, I'm not sure how it's going to be done.

src/modules/external_services/external_ip.py
RaitoBezarius: Alright, I'll perform those last changes tomorrow then and fix conflicts

src/modules/external_services/external_ip.py
iYehuda: In that case we can build the failover list with his preference first and append the original list, filtered:
```python
failover_list = [preference] + [service for service in services if service != preference]
```

danielsagi:
Closing.. a lot of those changes are not longer relevant. changing cloud discovery is  already planned to be implemented differently. and we no longer rely on canhazip.com

PR #271, opened 2019-10-31
Add multiple subscription mechanism

### LINKED ISSUE #144: Enhancing subscribe mechanism
The hunter subscribe mechanism currently support subscribing to one or multiple event.
The problem arises when for example we want to run a hunter only if two specific events were published.

One thing we could do is add a new 'queue' dictionary, that will store the 'multiple subscriptions', and on each event that is being published that resides in this queue_dict, adding the event object to the queue. 
In the case that the list of event objects corresponds to the requirement of the publish, the events will be concatenated, and the hunter hook will execute. This logic would happen each time on the publish_event() method.

RaitoBezarius:
## Description
As per #144 ; this PR enables a `multiple_subscribe` decorator which waits for multiple events to be fired before the hook is executed.
It also supports multiple predicates for each event.

It passes the set of `events` directly to the hook.

I guess we can refine it a lot more to fit more use cases, but I didn't want to go overkill, I believe the greatest generalization would be full logic `@complex_subscribe(SomeEvent & (EventA | ~EventB))` for example, and we could feed predicates using a dict. (where `&` means AND, `|` means OR and `~` means NOT).

That sounds cool, but in practice, is there a need for it?

## Fixed Issues

Fixes #144.

## Contribution checklist

### DIFF
--- a/src/core/events/handler.py
+++ b/src/core/events/handler.py
@@ -47,7 +47,8 @@ class EventQueue(Queue, object):
 
         return wrapper
 
-    def multiple_subscribe(self, events, predicates=None):
+    # decorator wrapping for multiple subscriptions
+    def subscribe_many(self, events, predicates=None):
         def wrapper(hook):
             self.subscribe_events(events, hook=hook, predicates=predicates)
             return hook
@@ -122,7 +123,7 @@ class EventQueue(Queue, object):
             for event, predicate in zip(events, predicates):
                 self.multi_hooks[event].append((hook, predicate))
 
-            self.hook_dependencies[hook] = set(events)
+            self.hook_dependencies[hook] = frozenset(events)
 
     def apply_filters(self, event):
         # if filters are subscribed, apply them on the event
--- a/src/core/events/handler.py
+++ b/src/core/events/handler.py
@@ -23,6 +23,9 @@ class EventQueue(Queue, object):
 
         self.hooks = defaultdict(list)
         self.filters = defaultdict(list)
+        self.multi_hooks = defaultdict(list) # for hooks which listens for multiple events.
+        self.hook_dependencies = defaultdict(set) # dependencies which were defined for a given hook.
+        self.hook_fulfilled_deps = defaultdict(set) # keep track of already fulfilled dependencies.
         self.running = True
         self.workers = list()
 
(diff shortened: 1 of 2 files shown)


tests/core/test_subscribe.py
```diff
@@ -26,25 +26,38 @@ def __init__(self, event):
...
-    
+
+    # testing the multiple subscription mechanism
+    handler.publish_event(OnceOnlyEvent())
+
+    time.sleep(0.02)
+    # SmartHunter and OnceHunter should run only.
+    assert counter == 2
+    counter = 0
+
```
iYehuda: I think this code deserves its own test (i.e, "test_subscribe_many").
The current test is already overloaded with "subscribe" and "subscribe_once" testing, so you can split that too (please).

tests/core/test_subscribe.py
```diff
@@ -26,25 +26,38 @@ def __init__(self, event):
...
     handler.publish_event(RegularEvent())
-    
+
     time.sleep(0.02)
     assert counter == 3
     counter = 0
-    
+
+    # testing the multiple subscription mechanism
+    handler.publish_event(OnceOnlyEvent())
```
iYehuda: Please add test case of subscriber being triggered twice.

src/core/events/handler.py
```diff
@@ -44,6 +47,13 @@ def wrapper(hook):
 
         return wrapper
 
+    def multiple_subscribe(self, events, predicates=None):
```
iYehuda: Rename to `subscribe_many` to align with `subscribe_once`

iYehuda:
I'd like this PR to get another pair of eyes on it since it's a little complicated.
@itaysk can you review it?

RaitoBezarius:
I addressed some comments, but it's not passing tests anymore, I have to see why. I'm just worrying about the fact of splitting the test functions might have created some implicit global state that I don't see. :/

RaitoBezarius:
Also, it brings some question.

If I have some hunter depending on A & B

If I publish A, A, A, …, A, B, B, B, …, B

How many times should the hunter be called?

iYehuda:
> Also, it brings some question.
> 
> If I have some hunter depending on A & B
> 
> If I publish A, A, A, …, A, B, B, B, …, B
> 
> How many times should the hunter be called?

That's a good point.
Let's examine the use case with an example:
* Wer'e running on Azure (so AzureMetadataAPI event is published)
* There are 2 vulnerable kubelet APIs (so SecureKubeletEvent is published twice)
Assume there's a new vulnerability in azure that depends on the former combination, we would like to get it triggered for every kubelet found but we need only one metadata api as prerequisite.

It is possible to have a case which additional context should be included in triggering, so supplying a predicate for two events can be done to filter context.
In such case, you might have event A published twice and event B published twice as well, but a registered hunter C might need to have the second A and first B.
This way, you'll need to trigger C for each combination of A and B events and let it filter.

danielsagi:
Closing due to inactivity

Issue #254, opened 2019-10-28
Expose VID on all output types


itaysk:
The VID is currently exposed only for "plain" output (meaning, the terminal output).
We should add it to other types like json/yaml.
Unlike the plain output, where it didn't make sense to add the full url to the report for usability concerns, in the other outputs which are machine readable we should include the url.

(this is also requested for Aqua CSP integration)

RaitoBezarius:
Hi @itaysk 

AFAIK, VID is exposed in JSON/YAML, I can see a `vid` field on `master` version.
But I agree on full URL, I'll send a PR right away if it's okay.

RaitoBezarius:
Well, it's not obvious what would be the full URL because I don't see any hosted docs somewhere, @itaysk — also do we want to hardcode a base URL?

itaysk:
Yes the VID is already there, I didn't notice. Please ignore this issue for now.

PR #275, opened 2019-11-10
Refactor reporters and add KB URL to reports

itaysk:
## Description
Refactor reporters to remove code duplication (moved shared logic to BaseReport and rebased YAMLReporter on JSONReporter)
Also add KB URL to report output.

## Fixed Issues
fix: https://github.com/aquasecurity/kube-hunter/issues/254


src/modules/report/base.py
```diff
@@ -1,5 +1,7 @@
 from .collector import services, vulnerabilities, hunters, services_lock, vulnerabilities_lock
 from src.core.types import Discovery
+from __main__ import config
```
iYehuda: Note that once [this PR](https://github.com/aquasecurity/kube-hunter/pull/264) is merged, we will stop using `__main__` imports and fetch configuration from separate module.

PR #272, opened 2019-11-02
Packaging Kube-Hunter for PyPi

vipulgupta2048:
Signed-off-by: Vipul Gupta (@vipulgupta2048) <vipulgupta2048@gmail.com>

## Description
PR to solve issue #185, package the Kube-Hunter project, upload it to PyPi and make it viable for it to be installed through pip for our users, on all platforms. Right now, it has been deployed to TestPyPi here https://test.pypi.org/project/kube-hunter/0.1.0/ 
Can share the credentials with the maintainers if you wish to get access. 

## Things to work out 
- ~Making Kube-Hunter work as a pip package.~
- ~Verify sample metadata filled in setup.py and get that approved.~

I would recommend going the CLI way with Kube-Hunter, where users can use the command for example for active hunting with `kube-hunter --remote some.domain.com --active`
That way, we won't need to do a lot of changes and it's quite easier for users to learn and take up. Let me know what you think folks?

## Contribution Guidelines
Yes, I read them :heavy_check_mark: 

## Fixed Issues
#185

PR #283, opened 2019-12-03
Refactor configuration 

### LINKED ISSUE #234: Profiling is impossible on the script

Currently, due to the usage of `__main__` to import the `config` object from ArgParse, it is impossible to use external scripts that can instrument the main script (because it changes the entrypoint script).

Making it impossible to use: `cProfile`, `memory_profiler` for instance.

A solution is to create another file `conf.py` and add a global `config` object in it which is initialized from the `kube-hunter.py` script and every other parts of the script can depend on the `conf` module rather than `__main__` (which violates PEP8 BTW).

## Why is this needed

Profiling the script helps developers to debug performance/memory issues, which is critical for a security script.

iYehuda:
## Description
Moved configuration into a separate package and break dependencies of `__main__` module.
This way we can run external tools on kube-hunter (e.g., pytest, profiling, etc.).
runtest.py may now be deprecated and removed.

## Contribution Guidelines

## Fixed Issues
Fixes #234 

## Contribution checklist
 
## Notes
This PR was initiated by @RaitoBezarius, thank you for your work!


runtest.py
```diff
@@ -1,27 +1,12 @@
...
-parser.add_argument('--remote', nargs='+', metavar="HOST", default=list(), help="one or more remote ip/dns to hunt")
-parser.add_argument('--active', action="store_true", help="enables active hunting")
-parser.add_argument('--log', type=str, metavar="LOGLEVEL", default='INFO', help="set log level, options are: debug, info, warn, none")
-parser.add_argument('--report', type=str, default='plain', help="set report type, options are: plain, yaml")
-parser.add_argument('--statistics', action="store_true", help="set hunting statistics")
-
-config = parser.parse_args()
+#!/usr/bin/env python3
 
+import argparse
```
itaysk: shouldn't this import conf instead of argparse?

kube-hunter.py
```diff
@@ -1,63 +1,55 @@
...
-
-import plugins
-
-config = parser.parse_args()
-
-try:
-    loglevel = getattr(logging, config.log.upper())
-except:
-    pass
+# TODO: move log level parsing to conf module
```
itaysk: should we create a new issue to track the conf refactor and add these TODOs there? I understand you don't want to block this PR on all other little refactoring but we should be able to better track it

iYehuda: I believe it is a good idea

runtest.py
iYehuda: It should be deleted entirely, but considering this file is made irrelevant by #272 we can ignore that.

kube-hunter.py
itaysk: created https://github.com/aquasecurity/kube-hunter/issues/289. please remove the comments before merging

PR #302, opened 2020-02-15
Add PyInstaller build

### LINKED ISSUE #301: Standard build with PyInstaller
Add support for PyInstaller build.
That is not trivial since there are dynamic imports in kube-hunter code base, which should be treated individually when building \w PyInstaller.

## Why is this needed
There are kinds of integrations which cannot have python exists.
We can use PyInstaller to "freeze" kube-hunter to a single executable binary.
Going further we can use tools such as [staticx](https://github.com/JonathonReinhart/staticx) to generate a standalone statically linked binary, thus being portable to non glibc environments. 

Related to #290

iYehuda:
## Description
Add standard method for building with PyInstaller.
Use staticx to generate a single static binary.
PyInstaller and staticx builds given as Makefile targets `pyinstaller` and `pyinstaller_static` accordingly.

## Contribution Guidelines

## Fixed Issues
Resolves #301

## Contribution checklist
 
## Notes
No functionality added, therefore, there's nothing to cover with tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,8 @@ var/
 *.egg-info/
 .installed.cfg
 *.egg
+*.spec
+.eggs
 
 # Directory Cache Files
 .DS_Store
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,10 @@ deps:
 lint:
 	flake8 $(SRC)
 
+.PHONY: test
+test:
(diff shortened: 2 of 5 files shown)


itaysk:
I'm unclear about the motivation for this pr, commented in the associated issue.

PR #303, opened 2020-02-16
Fix Dockerfile build

### LINKED ISSUE #300: Docker image build fails
### What happened
Running `docker build .` fails with following error (which I think is something to do with the new PyPi packaging)

```
Step 5/12 : RUN pip install -r /kube-hunter/requirements.txt -t /kube-hunter
 ---> Running in 0271caccbe33
ERROR: File "setup.py" not found. Directory cannot be installed in editable mode: /kube-hunter
WARNING: You are using pip version 19.3.1; however, version 20.0.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
The command '/bin/sh -c pip install -r /kube-hunter/requirements.txt -t /kube-hunter' returned a non-zero code: 1
```

### Expected behavior

Successful docker image build

iYehuda:
## Description
The Docker build used a 2-step installation of requirements
and application.
This was broke by #272.

## Contribution Guidelines

## Fixed Issues

Fixes #300

### BEFORE
`docker build .` failed

### AFTER
`docker build .` works

## Contribution checklist
 
## Notes
Changed build file. No code to test.


.dockerignore
```diff
@@ -0,0 +1,4 @@
+*.png
+tests/
+docs/
+.github/
```
invidian: Missing newline.

Dockerfile
```diff
@@ -1,25 +1,16 @@
...
     tcpdump \
     build-base \
-    ebtables
+    ebtables \
+    make \
+    git && \
+    apk upgrade --no-cache
 
 WORKDIR /kube-hunter
-COPY ./requirements.txt /kube-hunter/.
```
invidian: Hm, it would be nice to have `pip install` layer cached, otherwise any code change triggers downloading the packages again, so rebuilding image takes ~30-40 sec.

.dockerignore
iYehuda: Done

Dockerfile
iYehuda: Good point.
Since requirements file doesn't include dependencies directly, we need another way of installing them and preserve check.
I can use `setup.py` custom command to list dependencies from `setup.cfg` and use it in Dockerfile

iYehuda: Done

iYehuda:
> 1. instead of the setuptools custom command, why not just run `pip install .`, which will automatically discover packages in `setup.cfg`?
> 2. why are we using `--editable` with pip install?

1. In order to use docker build cache properly, we need to separate dependencies installation apart from kube-hunter installation. Using `pip install .` installs both so it can't be split into two steps.
2. `--editable` is meant mostly for developing/debugging purposes. The `install` make target shouldn't use it.

PR #325, opened 2020-03-22
Optimize Cloud Discovery

### LINKED ISSUE #324: Optimize NewHostEvent get_cloud()
Right now in HostDiscovery.execute() we try to map out a cidr and assign a cloud to it, in the code only the first IP of a cidr is sent to the get_cloud() func(for example out of 192.168.0.1/24, only 192.168.0.1 is sent)
while we ignore a few facts:
1. The first IP might not exist in the cloud, but another might exist.
2. The first IP might exist in Azure, but another may belongs to other cloud vendor (if subnet is big enough).

So if we scan all IP's that should solve the problem, but then we have a performance issue, because we will scan each and every IP without even knowing if theres a VM behind it.

## Why is this needed
It's a feature in kube-hunter that currently isn't working right, and it's causing some hunters to not deliver what they should do (aks.py AzureSpnHunter for example)

## What can it look like
We can use cached-property pip package, so that when we acually access the event.cloud property it will call up the get_cloud function.
So the event.cloud property will act as a lazy load - only when we access it, it will try to obtain it.

mormamn:
## Description
Switched get_cloud() to a lazy load, using cached_property.
Now only when you access event.cloud it will load what cloud it is.

## Contribution Guidelines

## Fixed Issues
Resolves #324 

Output shouldnt be different

## Contribution checklist
 - [X] I have read the Contributing Guidelines.
 - [X] The commits refer to an active issue in the repository.
 - [X] I have added automated testing to cover this case.

### DIFF
--- a/kube_hunter/core/events/types/common.py
+++ b/kube_hunter/core/events/types/common.py
@@ -4,6 +4,7 @@ import requests
 import logging
 
 from kube_hunter.core.types import InformationDisclosure, DenialOfService, RemoteCodeExec, IdentityTheft, PrivilegeEscalation, AccessRisk, UnauthenticatedAccess, KubernetesCluster
+from kube_hunter.conf import config
 
 logger = logging.getLogger(__name__)
 
@@ -135,16 +136,11 @@ class NewHostEvent(Event):
         try:
             logger.debug("Checking whether the cluster is deployed on azure's cloud")
             # Leverage 3rd tool https://github.com/blrchen/AzureSpeed for Azure cloud ip detection
-            metadata = requests.get(f"https://api.azurespeed.com/api/region?ipOrUrl={self.host}").text
+            return requests.get(f"https://api.azurespeed.com/api/region?ipOrUrl={self.host}", timeout=config.network_timeout).json()["cloud"]
         except requests.ConnectionError:
             logger.info(f"Failed to connect cloud type service", exc_info=True)
-            return
         except Exception:
             logger.warning(f"Unable to check cloud of {self.host}", exc_info=True)
-        if "cloud" in metadata:
-            return json.loads(metadata)["cloud"]
-        else:
-            return "NoCloud"
         
     def __str__(self):
         return str(self.host)
--- a/kube_hunter/core/events/types/common.py
+++ b/kube_hunter/core/events/types/common.py
@@ -2,7 +2,6 @@ import threading
 import json
 import requests
 import logging
-from cached_property import cached_property
 
 from kube_hunter.core.types import InformationDisclosure, DenialOfService, RemoteCodeExec, IdentityTheft, PrivilegeEscalation, AccessRisk, UnauthenticatedAccess, KubernetesCluster
 
@@ -126,7 +125,7 @@ class NewHostEvent(Event):
             self.event_id = event_id_count
             event_id_count += 1
 
-    @cached_property
+    @property
     def cloud(self):
         if not self.cloud_type:
             self.cloud_type = self.get_cloud()
@@ -144,6 +143,8 @@ class NewHostEvent(Event):
             logger.warning(f"Unable to check cloud of {self.host}", exc_info=True)
         if "cloud" in metadata:
             return json.loads(metadata)["cloud"]
+        else:
+            return "NoCloud"
         
     def __str__(self):
         return str(self.host)
--- a/kube_hunter/core/events/types/common.py
+++ b/kube_hunter/core/events/types/common.py
@@ -142,6 +142,7 @@ class NewHostEvent(Event):
             logger.info(f"Failed to connect cloud type service", exc_info=True)
         except Exception:
             logger.warning(f"Unable to check cloud of {self.host}", exc_info=True)
+        return "NoCloud"
         
     def __str__(self):
         return str(self.host)
--- a/kube_hunter/core/events/types/common.py
+++ b/kube_hunter/core/events/types/common.py
@@ -1,7 +1,12 @@
 import threading
+import json
+import requests
+import logging
+from cached_property import cached_property
 
 from kube_hunter.core.types import InformationDisclosure, DenialOfService, RemoteCodeExec, IdentityTheft, PrivilegeEscalation, AccessRisk, UnauthenticatedAccess, KubernetesCluster
 
+logger = logging.getLogger(__name__)
 
 class EventFilterBase(object):
     def __init__(self, event):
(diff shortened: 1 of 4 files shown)


kube_hunter/core/events/types/common.py
```diff
@@ -115,12 +120,31 @@ class NewHostEvent(Event):
...
         global event_id_count
         self.host = host
-        self.cloud = cloud
+        self.cloud_type = cloud
 
         with event_id_count_lock:
             self.event_id = event_id_count
             event_id_count += 1
 
+    @cached_property
```
iYehuda: Why do you use cached property?
It looks like you implement cache here

tests/core/test_cloud.py
```diff
@@ -0,0 +1,18 @@
...
+# Testing if it doesn't try to run get_cloud if the cloud type is already set.
+# get_cloud(1.2.3.4) will result with an error
+def test_presetcloud():
+    expcted = "AWS"
+    hostEvent = NewHostEvent(host="1.2.3.4", cloud=expcted)
+    assert expcted == hostEvent.cloud
+
+
+def test_getcloud():
+    AZURE_SERVER = "52.224.188.147" # this is portal.azure.com DNS record
```
iYehuda: I think it will be better to use a mock answer (using requests_mock) and not rely on this data to be persistent.

kube_hunter/core/events/types/common.py
mormamn: The cached property does help here, in case get_cloud() function didn't find a cloud so event.cloud_type remains None.
if it wasn't here every time event.cloud was accessed get_cloud() will be called, which isn't what we wanted

iYehuda: Your'e right but it seems that the case were cloud_type is defined never happens (please verify me).
In that case, we can drop this parameter from the constructor and then change the `cloud` property to the `get_cloud` function (with `cached_property` decorator).

mormamn: Looking at it now, i can see there are 2 places that try to set cloud_type 
they are under FromPodHostDiscovery, 2 functions called 
- azure_metadata_discovery = which returns hard-coded "Azure", we can remove that..
- traceroute_discovery = which returns pod external ip (?), we can also remove that since it's not a cloud_type..

if we do that, we can definitely do what u suggested

iYehuda: We can drop the usage of `@cached_property` if we return `"NoCloud"` from `get_cloud` when cloud is not found, instead of just returning `None`.
It looks like that the output from `traceroute_discovery` should go through the `get_cloud` function.
The `azure_metadata_discovery` return value is fine, although I believe that there should be a broader logic for dealing with cloud metadata server and specifically detecting cloud type from it, but it's out of scope.

kube_hunter/core/events/types/common.py
```diff
@@ -115,12 +119,33 @@ class NewHostEvent(Event):
...
+    def cloud(self):
+        if not self.cloud_type:
+            self.cloud_type = self.get_cloud()
+        return self.cloud_type
+    
+    def get_cloud(self):
+        try:
+            logger.debug("Checking whether the cluster is deployed on azure's cloud")
+            # Leverage 3rd tool https://github.com/blrchen/AzureSpeed for Azure cloud ip detection
+            metadata = requests.get(f"https://api.azurespeed.com/api/region?ipOrUrl={self.host}").text
```
iYehuda:
```suggestion
            metadata = requests.get(f"https://api.azurespeed.com/api/region?ipOrUrl={self.host}").json()
```

kube_hunter/core/events/types/common.py
```diff
@@ -115,12 +119,33 @@ class NewHostEvent(Event):
...
+        except requests.ConnectionError:
+            logger.info(f"Failed to connect cloud type service", exc_info=True)
+            return
```
iYehuda:
```suggestion
```

kube_hunter/core/events/types/common.py
```diff
@@ -115,12 +119,33 @@ class NewHostEvent(Event):
...
+        except Exception:
+            logger.warning(f"Unable to check cloud of {self.host}", exc_info=True)
+        if "cloud" in metadata:
+            return json.loads(metadata)["cloud"]
+        else:
+            return "NoCloud"
```
iYehuda:
```suggestion
        return "NoCloud"
```

kube_hunter/core/events/types/common.py
iYehuda: Also add a timeout from config.network_timeout

kube_hunter/modules/discovery/hosts.py
```diff
@@ -168,14 +155,13 @@ def execute(self):
             except ValueError as e:
                 logger.exception(f"Unable to parse CIDR \"{config.cidr}\"")
                 return
-            cloud = HostDiscoveryHelpers.get_cloud(ip)
             for ip in HostDiscoveryHelpers.generate_subnet(ip, sn=sn):
-                self.publish_event(NewHostEvent(host=ip, cloud=cloud))
+                self.publish_event(NewHostEvent(host=ip))
```
iYehuda: Why did you remove the `cloud` parameter?

setup.cfg
```diff
@@ -35,6 +35,7 @@ install_requires =
     PrettyTable
     urllib3>=1.24.3
     ruamel.yaml
+    cached-property
```
iYehuda: Not needed anymore

tests/core/test_cloud.py
```diff
@@ -0,0 +1,31 @@
...
+
+def test_getcloud():
+    fake_host = "1.2.3.4"
+    expected_cloud = "Azure"
+    result = {
+        "cloud": expected_cloud,
+        "regionId": "europenorth",
+        "region":"North Europe",
+        "location":"Ireland",
+        "ipAddress": fake_host
```
iYehuda: These values are not needed for the test
```suggestion
```

kube_hunter/modules/discovery/hosts.py
mormamn: this was the whole point of the issue, we don't need to check cloud type for each IP, we will only check it later when we access event.cloud

kube_hunter/core/events/types/common.py
```diff
@@ -115,12 +120,28 @@ class NewHostEvent(Event):
...
+    def cloud(self):
+        if not self.cloud_type:
+            self.cloud_type = self.get_cloud()
+        return self.cloud_type
+    
+    def get_cloud(self):
+        try:
+            logger.debug("Checking whether the cluster is deployed on azure's cloud")
+            # Leverage 3rd tool https://github.com/blrchen/AzureSpeed for Azure cloud ip detection
+            return requests.get(f"https://api.azurespeed.com/api/region?ipOrUrl={self.host}", timeout=config.network_timeout).json()["cloud"]
```
iYehuda: AzureSpeed can return `null` as the JSON value for `"cloud"`. In that case, return `"NoCloud"` rather than `None`.
```suggestion
            result = requests.get(f"https://api.azurespeed.com/api/region?ipOrUrl={self.host}", timeout=config.network_timeout).json()
            return result["cloud"] or "NoCloud"
```

kube_hunter/core/events/types/common.py
```diff
@@ -115,12 +120,28 @@ class NewHostEvent(Event):
...
+        except requests.ConnectionError:
+            logger.info(f"Failed to connect cloud type service", exc_info=True)
+        except Exception:
+            logger.warning(f"Unable to check cloud of {self.host}", exc_info=True)
```
iYehuda:
```suggestion
            logger.warning(f"Unable to check cloud of {self.host}", exc_info=True)
        return "NoCloud"
```

PR #347, opened 2020-04-29
docs: add requirement for setup tools-scm

lizrice:
Also assume pip3


iYehuda:
Wouldn't it be better to add setuptools-scm to `requirements.txt`?

lizrice:
I don't know, you tell me!

PR #360, opened 2020-06-26
Minor hunting bug fixes

danielsagi:
## Description
This PR fixes some typo issues and general misuses of function, which made some hunters fail silently.

## Contribution checklist

### DIFF
--- a/kube_hunter/modules/hunting/etcd.py
+++ b/kube_hunter/modules/hunting/etcd.py
@@ -121,7 +121,9 @@ class EtcdRemoteAccess(Hunter):
         logger.debug(f"{self.event.host} Passive hunter is attempting to read etcd keys remotely")
         try:
             r = requests.get(
-                f"{self.event.protocol}://{self.event.host}:{ETCD_PORT}/v2/keys", verify=False, timeout=config.network_timeout,
+                f"{self.event.protocol}://{self.event.host}:{ETCD_PORT}/v2/keys",
+                verify=False,
+                timeout=config.network_timeout,
             )
             self.keys_evidence = r.content if r.status_code == 200 and r.content != "" else False
             return self.keys_evidence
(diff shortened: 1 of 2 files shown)


kube_hunter/modules/hunting/kubelet.py
```diff
@@ -244,7 +244,7 @@ class DebugHandlers:
         """ all methods will return the handler name if successful """
 
         def __init__(self, path, pod, session=None):
-            self.path = path
+            self.path = path + "/" if not path.endswith("/") else ""
```
lizrice, line 247: Hmm...I think this is going to leave path as an empty string if it does end with /
```
>>> p = "hello/"
>>> p = p + "/" if not p.endswith("/") else ""
>>> p
''
```

Some extra brackets will fix it
```
>>> p = "hello/"
>>> p = p + ("/" if not p.endswith("/") else "")
>>> p
'hello/'
```

kube_hunter/modules/hunting/etcd.py
```diff
@@ -88,9 +88,7 @@ def db_keys_write_access(self):
         data = {"value": "remotely written data"}
         try:
             r = requests.post(
-                f"{self.protocol}://{self.event.host}:{ETCD_PORT}/v2/keys/message",
-                data=data,
-                timeout=config.network_timeout,
+                f"https://{self.event.host}:{ETCD_PORT}/v2/keys/message", data=data, timeout=config.network_timeout,
```
lizrice, line 91: Should this be `{self.event.protocol}`? I didn't check but I'm wondering if OpenPortEvent populates this. (In practice I would imagine any self-respecting deployment will use https anyway, so it's not a big deal, but just wondering about it)

kube_hunter/modules/hunting/kubelet.py
danielsagi: Youre right!

kube_hunter/modules/hunting/etcd.py
danielsagi: @lizrice 
Open port event does not populate it... but I changed it to save the "https" in the event, in any case if we want to use it in other event later on.

PR #362, opened 2020-08-07
Added multistage build for Dockerfile

danielsagi:
## Description
Until now we used alot of unnecessary alpine packages in the final image.
Now we only use the site-packages directory and the kube-hunter installed binary on a python alpine base image

## Fixed Issues

This removes vulnerabilities found in kube-hunter image. as well as reduce size from 300MB~ to 77MB!

## Contribution checklist
 
## Notes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,4 +16,8 @@ RUN make deps
 COPY . .
 RUN make install
 
+FROM python:3.8-alpine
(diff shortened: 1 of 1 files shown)


danielsagi:
@itaysk ,
Right, historically we installed them to remove warnings when loading scapy. in my tests I actually haven't seen warnings. 
But I added them again, seems to me that we will need them. the important part is removing the build packages.

PR #371, opened 2020-09-07
update gitignore

itaysk:
## Fixed Issues

add `pip-wheel-metadata` to the .gitignore

## Contribution checklist


evrardjp (commented):
How did you get that file? I don't see it after any make/python command in this project.

However, I understand why we have it in .gitignore, if we already have the .egg* present in gitignore. I just have the pip-wheel-metadata in my global gitignore list.

itaysk:
> How did you get that file? 
I just did `make deps` and `make dev-deps`

I saw that this was fixed in pip 20.2.2 which will use a temp dir instead, but I thought it would be better to add it to .getignore anyway

evrardjp:
Yes indeed, I am not against that :) 

I just thought it was worth a message in the commit message to explain the full context.

danielsagi (approved):
lgtm

Issue #370, opened 2020-09-06
Azure AzureSpnHunter is not working as expected


itaysk:
Azure AzureSpnHunter is not finding pods with access to /etc/kubernetes/azure.json
The code had a couple issues:
1. it checks the container mountPath, which only says about how the container sees the mount, not what is being mounted
2. it doesn't check the volume type. only hostPath volumes should be considered
I have a PR to fix that

PR #372, opened 2020-09-07
fix azure spn hunter

itaysk:
Fixes https://github.com/aquasecurity/kube-hunter/issues/370


kube_hunter/modules/hunting/aks.py
```diff
@@ -46,11 +46,16 @@ def get_key_container(self):
             logger.debug("failed getting pod info")
         else:
             pods_data = r.json().get("items", [])
+            suspecious_volume_names = []
```
evrardjp, line 49: nit: isn't this spelled suspicious?

kube_hunter/modules/hunting/aks.py
```diff
@@ -46,11 +46,16 @@ def get_key_container(self):
...
             for pod_data in pods_data:
+                for volume in pod_data["spec"]["volumes"]:
```
evrardjp, line 51: nit: Not sure whether the spec can _not_ have volumes. If that's the case, isn't it worth writing as `for volume in pod_data['spec'].get('volumes',[])` ? Or maybe have a try: except: KeyError?

kube_hunter/modules/hunting/aks.py
itaysk: yes thanks

kube_hunter/modules/hunting/aks.py
itaysk: you are absolutely right I forgot to check for failure cases. Thank you for spotting that.

danielsagi (approved):
lgtm

PR #376, opened 2020-10-03
Update __init__.py

bhavik9988:
accept this pull request


danielsagi:
Hi,
Thanks for you contribution! 
This PR adds no value to kube-hunter so we will not merge it. 
However, don't let this discourage you for making future PRs, I believe that with some time and effort you can dive into the code and make some real contributions!

PR #378, opened 2020-10-03
Fixed typo "web site" -> "website"

Zerplin:


danielsagi:
Hi, thanks for contributing to kube-hunter!
However, this change does not add real value to the project, so we will not merge it. This is distracting us from handling PRs that are adding features to the code/fixing bugs.
Don't let this discourage you though, I'm sure that with some time you could contribute something very valubale 💯

PR #387, opened 2020-10-12
Adding support for log redirection using --log-file option (#329)

### LINKED ISSUE #329: Log output redirection
Enable separation of report and log into two separate destinations.
Supply a configurable log output file.

## Why is this needed
In order to make kube-hunter automation and human friendly, we should enable separation of logs and reports in a convenient way.
Right now, log messages are being printed to the standard error and the report is written into the standard output.
If you want to separate them apart, you'll need to use shell redirection (e.g., `2> kube-hunter.log`).
We can save the user from using shell output redirection and supply a parameter for log file path (`--log-file`).

milindchawre:
## Description
Fixes #329 
Added `--log-file` option to redirect log output to a user specified file.

## Contribution Guidelines

## Fixed Issues

Please mention any issues fixed in the PR by referencing it properly in the commit message.
As per the convention, use appropriate keywords such as `fixes`, `closes`, `resolves` to automatically refer the issue.
Please consult [official github documentation](https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords) for details.

Fixes #(329)

### BEFORE
No way to redirect log output to any file.

### AFTER
hunterlog1

## Contribution checklist
 
## Notes
Have modified the existing test case for logging. Not completely sure whether that is enough.

### DIFF
--- a/kube_hunter/conf/parser.py
+++ b/kube_hunter/conf/parser.py
@@ -57,6 +57,13 @@ def parser_add_arguments(parser):
     )
 
     parser.add_argument(
+        "--log-file",
+        type=str,
+        default=None,
+        help="Path to a log file to output all logs to",
+    )
+
+    parser.add_argument(
         "--report",
         type=str,
         default="plain",
--- a/kube_hunter/__main__.py
+++ b/kube_hunter/__main__.py
@@ -18,6 +18,7 @@ config = Config(
     cidr=args.cidr,
     include_patched_versions=args.include_patched_versions,
     interface=args.interface,
+    log_file=args.log_file,
     mapping=args.mapping,
     network_timeout=args.network_timeout,
     pod=args.pod,
@@ -25,7 +26,7 @@ config = Config(
     remote=args.remote,
     statistics=args.statistics,
 )
-setup_logger(args.log)
+setup_logger(args.log, args.log_file)
 set_config(config)
 
 # Running all other registered plugins before execution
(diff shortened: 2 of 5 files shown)


milindchawre:
ping @iYehuda

kube_hunter/conf/parser.py
```diff
@@ -56,6 +56,13 @@ def parser_add_arguments(parser):
         help="Set log level, options are: debug, info, warn, none",
     )
 
+    parser.add_argument(
+        "--log-file",
+        type=str,
+        default=None,
+        help="Set logfile",
```
danielsagi, line 63: It would be better to write something more indicative. like: "Path to a log file to output all logs to"

danielsagi (changes requested):
Looks really good!
Thanks for your contribution. I just had a minor comment, after that I think we will be able to continue :)

kube_hunter/conf/parser.py
milindchawre: Done.

danielsagi (approved):

NickYan7:
Nice job!
And how can I redirect the result when I use docker to deploy the kube-hunter?
Such as I execute `docker run -it --rm --network host xxx/kube-hunter`, the `--log-file` is not work...
If I execute `docker run -it --rm --network host xxx/kube-hunter > ./result.txt`, it won't accept the any parameter so that I can't use kube-hunter by option 2 (Interface Scaning).

I also tried `echo 2 | docker run -it --rm --network host xxx/kube-hunter > ./scan_result.txt`, it is not work although... 😂
Screenshot 2021-08-24 下午3 12 15

Check this please, thanks a lot

@milindchawre

PR #388, opened 2020-10-15
Updated requirements.txt

imakish:
Added Requirements.


danielsagi:
Hi @imakish 
Thanks for you contribution!
But If you take a look at requirements-dev.txt you can see we already have this,
Don't let this discourage you though, with some time you can dive to the project really quick and I'm sure you can find ways to learn from it and add some value.
We certainly would love to see that.

PR #409, opened 2020-11-15
More intuitive message when ProveSystemLogs fails

### LINKED ISSUE #408: Proving system logs outputs misleading evidence
When the `/logs` endpoint is exposed, and an active scan is running, we try to parse the file `/var/log/audit/audit.log`
This file down not necessarily exist. for now we at least need to specify that we did not manage to prove this in the evidence.

Next step is implementing another way of proving this in a more versatile way

danielsagi:
## Description
ProveSystemLogs was misleading in certain cases on it's audit log parsing.
Now when the file does not exist, we output a corresponding message in the evidence

## Fixed Issues
fixes #408 

## Contribution checklist

### DIFF
--- a/kube_hunter/modules/hunting/kubelet.py
+++ b/kube_hunter/modules/hunting/kubelet.py
@@ -1147,8 +1147,8 @@ class ProveSystemLogs(ActiveHunter):
             timeout=config.network_timeout,
         )
 
-        #TODO: add more methods for proving system logs
-        if audit_logs.status_code == requests.status_codes.codes.OK:    
+        # TODO: add more methods for proving system logs
+        if audit_logs.status_code == requests.status_codes.codes.OK:
             logger.debug(f"Audit log of host {self.event.host}: {audit_logs.text[:10]}")
             # iterating over proctitles and converting them into readable strings
             proctitles = []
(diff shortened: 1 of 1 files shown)


kube_hunter/modules/hunting/kubelet.py
```diff
@@ -1145,11 +1145,16 @@ def execute(self):
...
-        self.event.evidence = f"audit log: {proctitles}"
+        )
+
+        # TODO: add more methods for proving system logs
+        if audit_logs.status_code == requests.status_codes.codes.OK:
+            logger.debug(f"Audit log of host {self.event.host}: {audit_logs.text[:10]}")
+            # iterating over proctitles and converting them into readable strings
+            proctitles = []
+            for proctitle in re.findall(r"proctitle=(\w+)", audit_logs.text):
+                proctitles.append(bytes.fromhex(proctitle).decode("utf-8").replace("\x00", " "))
```
lizrice, line 1156: Just a thought, I wonder if we still need this now we're using logs.text rather than just logs?

lizrice (approved):
I asked a question you might want to look at, and also note that I haven't tested so please make sure you have :-) but otherwise lgtm

danielsagi (commented):
Commented, and yes. Tested!

lizrice (approved):

PR #411, opened 2020-11-15
Fixed false positive on test_run_handler

### LINKED ISSUE #410: False positives on ExposedRunHandler
Until now we relied on a Method not allowed response status code (405) when querying for a dummy request to the `/run` endpoint.
While it is a sign that we passed Authorization, it does not mean that the endpoint is enabled. When debug handlers are disabled the status code remains the same.
This resulted in a False positive, so even when debug handlers are disabled, we raised an `ExposedRunHandler`

The best way to check this is by getting the error message `"pod does not exist"`

danielsagi:
## Description
As described in #410 
We now compare the resulted text body to a hardcoded error message `pod does not exist` coming from kubelet.
By looking at the [source code of the kubelet](https://github.com/kubernetes/kubernetes/blob/f5abe26a198d845f0a3cd6bd13ab9a37adda90b5/pkg/kubelet/server/server.go#L811)
This looks like this is not gonna change in the long term

## Fixed Issues
Fixes #410 

## Contribution checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
     - "3.6"
     - "3.7"
     - "3.8"
+    - "3.9"
 install:
   - pip install -r requirements.txt
   - pip install -r requirements-dev.txt
--- a/docs/_kb/KHV005.md
+++ b/docs/_kb/KHV005.md
@@ -12,7 +12,7 @@ Kubernetes API was accessed with Pod Service Account or without Authentication (
 
 ## Remediation
 
-Secure acess to your Kubernetes API.
+Secure access to your Kubernetes API.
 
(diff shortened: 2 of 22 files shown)


lizrice (commented):
I'm wondering if we shouldn't look for the StatusNotFound (404) error code that gets returned along with that error message - is there a way for 404 to get returned in any other circumstances except that the client has passed authn/z but there is no matching pod?

danielsagi:
@lizrice 
So when we get this error message we dont get a 404 status, but rather a 405 as well.
This was the problem. also, 404 gets returned regardless of authorization/authentication. only when a specific endpoint does not exist entirely

danielsagi:
@lizrice oh I get what you were asking, Ill try maybe to send a post request, perhaps well get a 404 and can rely on that.

itaysk (approved):

Issue #413, opened 2020-11-27
Implement Port Forward Test Hunting


danielsagi:
## What would you like to be added

Adding an implementation for reporting an open `/portforward` debug handler
[](https://github.com/aquasecurity/kube-hunter/blob/14ca1b8bcef9822aeda2f6992450822bdc957e41/kube_hunter/modules/hunting/kubelet.py#L346)

## Why is this needed

Currently there is no indication for the existence of this handler.
The goal is adding a way of testing this is open, without actually making changes to the cluster. (Passive check)
Of course there is room for an additional active hunter that uses this :)

PR #399, opened 2020-10-29
Migrate tests to GitHub Actions

hugovk:
## Description
Please include a summary of the change and which issue is fixed. Also include relevant motivation and context. List any dependencies that are required for this change.

Run tests using GitHub Actions:

* It has 20 parallel runners compared to officially 5 on Travis, but nowadays often 4 or less

* Easy to test on Ubuntu 16.04, 18.04 and 20.04

* We can put lint in a separate build job, so if something fails it's obvious right away if the problem is linting or the tests, without clicking through and opening and scrolling the build logs

* pre-commit already does the lint check, it's simple and clean to use that on the CI too. A benefit is it uses the same pinned linter versions as local pre-commit use.

## Contribution Guidelines
Please Read through the [Contribution Guidelines](https://github.com/aquasecurity/kube-hunter/blob/master/CONTRIBUTING.md).

## Fixed Issues

Please mention any issues fixed in the PR by referencing it properly in the commit message.
As per the convention, use appropriate keywords such as `fixes`, `closes`, `resolves` to automatically refer the issue.
Please consult [official github documentation](https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords) for details.

Fixes https://github.com/aquasecurity/kube-hunter/issues/395

## "BEFORE" and "AFTER" output

To verify that the change works as desired, please include an output of terminal before and after the changes under headings "BEFORE" and "AFTER".

### BEFORE
Any Terminal Output Before Changes.

n/a

### AFTER
Any Terminal Output Before Changes.

n/a

Sample runs:

* https://github.com/hugovk/kube-hunter/actions/runs/336357687
* https://github.com/hugovk/kube-hunter/actions/runs/336357693

## Contribution checklist
 - [x] I have read the Contributing Guidelines.
 - [x] The commits refer to an active issue in the repository.
 - [x] I have added automated testing to cover this case.
 
## Notes
Please mention if you have not checked any of the above boxes.

PR #414, opened 2020-11-27
Fix lint comments

danielsagi:
## Description
Commented out test_port_forward function in kubelet hunting.
I created an issue #413 that concerns this, but we can't have a non complete function with flake8 running.

This is regarding moving our tests to github actions #399 , the flake8 in the test environment complains about this function.

## Contribution checklist


lizrice (approved):
ok

PR #430, opened 2021-01-12
Update README.md

franciosi:
Adding simple numbers to make the install documentation easier

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,9 +115,9 @@ Available dispatch methods are:
     * KUBEHUNTER_HTTP_DISPATCH_METHOD (defaults to: POST)
 
 ## Deployment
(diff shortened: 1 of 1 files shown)


danielsagi:
Hi @franciosi 
Thanks for your contribution!
However, I don't see how adding numbers in the readme makes difference in understanding it better..
I encourage you to add a contribution in the code :)

franciosi:
Hi @danielsagi

OK, it was just a simple improve. This is how normal we do with documentation, like [https://www.freebsd.org/doc/handbook/](https://www.freebsd.org/doc/handbook/). About code, yes, I plan to do it, as soon as I have more time. Very much involved with FOSS documentation and translations for now.

PR #431, opened 2021-01-14
deps: Update github pages dependencies

simar7:
## Description
Updated kramdown to patch for CVE-2020-14001.
Updated nokogiri to patch for GHSA-vr8q-g5c7-m54m.

It also updates github-pages dependency as it was needed to update the above dependencies. Also here's the new list of recommended dependency version for github-pages: https://pages.github.com/versions/

## Contribution checklist

On a side note, should we still maintain the docs here as we've moved AVD now and https://aquasecurity.github.io/kube-hunter/ is a 404. Maybe this PR isn't needed after all.

Just to be clear: AVD still uses the markdowns generated in https://github.com/aquasecurity/kube-hunter/tree/main/docs/_kb to build the content pages for itself. 

Thoughts?

Signed-off-by: Simarpreet Singh <simar@linux.com>


docs/Gemfile.lock
```diff
@@ -230,21 +238,27 @@ GEM
...
-    unicode-display_width (1.6.0)
+    unf (0.1.4)
+      unf_ext
+    unf_ext (0.0.7.7)
+    unicode-display_width (1.7.0)
+    zeitwerk (2.4.2)
 
 PLATFORMS
-  ruby
+  x86_64-darwin-19
```
itaysk, line 257: probably shouldn't be here? maybe `bundle lock --add-platform ...` will help

simar7: Ah yes - thanks

```bundle lock --add-platform ruby --remove-platform x86_64-darwin-19``` 

did the trick.

itaysk (approved):

PR #436, opened 2021-01-20
Document --quick option in README

mlevesquedion:
## Description
This change adds a bit of documentation concerning the `--quick` option. I did not create an issue beforehand as I believe that this is a "trivial documentation" change (as described in `CONTRIBUTING.md`).  This change was inspired by discussion on https://github.com/aquasecurity/starboard/issues/165.

## Contribution Guidelines

## Fixed Issues

N/A

N/A

## Contribution checklist
 - [x] The commits refer to an active issue in the repository. (N/A)
 - [x] I have added automated testing to cover this case. (N/A)
 
## Notes

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,10 @@ To specify interface scanning, you can use the `--interface` option (this will s
 To specify a specific CIDR to scan, use the `--cidr` option. Example:
 `kube-hunter --cidr 192.168.0.0/24`
 
+4. **Quick scanning**
+
+To limit subnet scanning to a `/24` CIDR, use the `--quick` option. This option only applies when running as a Pod in Azure.
+
 ### Active Hunting
 
 Active hunting is an option in which kube-hunter will exploit vulnerabilities it finds, to explore for further vulnerabilities.


danielsagi:
Hi @mlevesquedion
That's a really good idea!
quick scanning is a feature that solves a problem in azure scanning. I agree that this should be documented. But I believe we should create something like a `Q&A.md` or `ADVANCEDUSAGE.md`  and in there, start adding knowledge like that. 
Because it is not falling under a scanning method,  but rather falls under general knowledge.
Wdyt?

mlevesquedion:
Hi @danielsagi 

Thanks for taking a look at my PR! I agree that this might not belong in the `README`. On the other hand, I'm not sure it's worth creating an extra file just for this piece of documentation. Are there other things which currently aren't documented, but that we could put into such a `Q&A` or `ADVANCEDUSAGE` file?

danielsagi:
Hi @mlevesquedion 
Sorry  for the late reply, I agree. Maybe add an Advance Usage section in the original Readme file.

danielsagi:
closing, due to inactivity... I created a new PR for this #441

PR #478, opened 2021-08-26
modify error when cluster resources  unreachable

### LINKED ISSUE #477: Modify Error Message
Currently if a node is unreachable by the DNS or ip address in bare python installation of kube hunter it returns an error of Kube Hunter couldn't find any clusters . This error message can be modified to something like Host is inaccessible or cluster is inaccessible/unreachable. 
This message has to be modified https://github.com/aquasecurity/kube-hunter/blob/473e4fe2b57d3e91b91777de0ed65d13658ae9cb/kube_hunter/modules/report/plain.py#L45
This could help prevent issues like #404  and #475.

## Why is this needed
The error message currently implies that no cluster is found even though cluster is present  however it should imply that the cluster nodes are not reachable from externally.

DiptoChakrabarty:
## Description
Please include a summary of the change and which issue is fixed. Also include relevant motivation and context. List any dependencies that are required for this change.

Modifies the Error message when cluster resources are unreachable from kube hunter

## Contribution Guidelines

## Fixed Issues

Please mention any issues fixed in the PR by referencing it properly in the commit message.
As per the convention, use appropriate keywords such as `fixes`, `closes`, `resolves` to automatically refer the issue.
Please consult [official github documentation](https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords) for details.

Fixes #477 

### BEFORE
Any Terminal Output Before Changes.
Kube Hunter couldn't find any clusters

### AFTER
Any Terminal Output Before Changes.
Cluster Inaccessible from Kube Hunter

## Contribution checklist
 
## Notes
As tests are not required for this.

### DIFF
--- a/kube_hunter/core/events/handler.py
+++ b/kube_hunter/core/events/handler.py
@@ -137,10 +137,8 @@ class EventQueue(Queue):
         # Document that the hunter published a vulnerability (if it's indeed a vulnerability)
         # For statistics options
         self._increase_vuln_count(event, caller)
-
         # sets the event's parent to be it's publisher hunter.
         self._set_event_chain(event, caller)
-
         # applying filters on the event, before publishing it to subscribers.
         # if filter returned None, not proceeding to publish
         event = self.apply_filters(event)
@@ -224,8 +222,6 @@ class EventQueue(Queue):
         if caller:
             event.previous = caller.event
             event.hunter = caller.__class__
-            if "error" in caller.__dict__:
-                event.error = caller.error
 
     def _register_hunters(self, hook=None):
         """
--- a/kube_hunter/core/events/handler.py
+++ b/kube_hunter/core/events/handler.py
@@ -224,6 +224,8 @@ class EventQueue(Queue):
         if caller:
             event.previous = caller.event
             event.hunter = caller.__class__
+            if "error" in caller.__dict__:
+                event.error = caller.error
 
     def _register_hunters(self, hook=None):
         """
--- a/kube_hunter/core/events/types.py
+++ b/kube_hunter/core/events/types.py
@@ -32,6 +32,7 @@ class Event:
     def __init__(self):
         self.previous = None
         self.hunter = None
+        self.error = None
 
     # newest attribute gets selected first
     def __getattr__(self, name):
--- a/kube_hunter/core/events/types.py
+++ b/kube_hunter/core/events/types.py
@@ -190,11 +190,15 @@ class NewHostEvent(Event):
 
 
 class OpenPortEvent(Event):
-    def __init__(self, port):
+    def __init__(self, port=None,error="default"):
         self.port = port
+        self.error = error
 
     def __str__(self):
         return str(self.port)
+    
+    def __repr__(self):
+        return str(self.error)
 
     # Event's logical location to be used mainly for reports.
     def location(self):
--- a/kube_hunter/core/events/types.py
+++ b/kube_hunter/core/events/types.py
@@ -32,7 +32,7 @@ class Event:
     def __init__(self):
         self.previous = None
         self.hunter = None
-        self.error = None
+        #self.error = None
 
     # newest attribute gets selected first
     def __getattr__(self, name):
@@ -138,6 +138,12 @@ class Vulnerability:
     def get_severity(self):
         return self.severity.get(self.category, "low")
 
+class HuntError(Event):
+    def __init__(self,error):
+        self.error = error
+
+    def __str__(self):
+        return str(self.error)
 
 event_id_count_lock = threading.Lock()
 event_id_count = 0
--- a/kube_hunter/core/events/types.py
+++ b/kube_hunter/core/events/types.py
@@ -32,7 +32,6 @@ class Event:
     def __init__(self):
         self.previous = None
         self.hunter = None
-        #self.error = None
 
     # newest attribute gets selected first
     def __getattr__(self, name):
--- a/kube_hunter/core/events/types.py
+++ b/kube_hunter/core/events/types.py
@@ -138,6 +138,7 @@ class Vulnerability:
         return self.severity.get(self.category, "low")
 
 class HuntError(Event):
+    
     def __init__(self,error):
         self.error = error
 
--- a/kube_hunter/modules/discovery/apiserver.py
+++ b/kube_hunter/modules/discovery/apiserver.py
@@ -52,9 +52,13 @@ class ApiServiceDiscovery(Discovery):
     def execute(self):
         logger.debug(f"Attempting to discover an API service on {self.event.host}:{self.event.port}")
         protocols = ["http", "https"]
+        flag = True
         for protocol in protocols:
             if self.has_api_behaviour(protocol):
+                flag = False
                 self.publish_event(K8sApiService(protocol))
+        if flag:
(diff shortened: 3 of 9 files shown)


danielsagi:
Hi @DiptoChakrabarty 
Thats a really good change, I thought about changing this behavior. 
However, I believe in order for this error message to be more indicative, we need to add some mechanism to determine why kube hunter failed to discover the cluster.

The most common problems that cause kube hunter to fail are:

1. Ports are not open from where kube-hunter was ran

2. Ports are open, but kube-hunter could not identify a kubernetes service behind them

3. Failed dns resolution

4. The ip is not reachable (not very easy to distinguish from ports not open, because ping requests might be blocked anyway)

So if we were to change the message to be more indicative I want that to be by adding those 3 messages.

Do you think this is something you want to take on yourself?
This would be wonderful :)
I would help in the process with questions of course.

The general process of identifying the specific problem could be by checking the port open event thats fired if we found an open port.

DiptoChakrabarty:
Yes @danielsagi  I would like to try it out , I think the ports issue is visible when we run it with debug log flag so maybe can get some pointers from there Ill check it out

DiptoChakrabarty:
hey @danielsagi I am able to log out the errors could you  give me some pointers on how I could display these errors , I know that the final error finally comes from here https://github.com/aquasecurity/kube-hunter/blob/473e4fe2b57d3e91b91777de0ed65d13658ae9cb/kube_hunter/modules/report/plain.py#L16-L46 . How can I propagate the error here

DiptoChakrabarty:
Selection_138
Hey @danielsagi  I am able to pass errors around and getting the port not open message , however that is only because I do a publish_event of HuntFinished in the port discovery class  as huntfinished of main does not pickup the PortDiscovery as caller in the publish_event.

I was thinking how I could pass the PortDiscovery to teh HuntFinished of the main function which would give the required output.

Currently if a port or service is not discovered no event is published due to which SendFullReport  which picks up from HuntFinished does not have any details of the error , if there was a publish event type we could run which would directly go to Huntfinished I think it could solve the problem.

danielsagi:
Hi @DiptoChakrabarty 
Glad to see you working on this! Ill be free at the end of this week to help you, please be patient and well work on this together :)

DiptoChakrabarty:
> Hi @DiptoChakrabarty
> Glad to see you working on this! Ill be free at the end of this week to help you, please be patient and well work on this together :)

Okay Sure  :+1:

DiptoChakrabarty:
@danielsagi can you check the commits once , I am able to print the error and propagate it but stuck on the last step

danielsagi (changes requested):
I have looked at your code. sorry it took too long.
It was a really creative way of how you did the report collection, binding errors to hunters and propagating on events.  
Though I believe in order to solve the last issue we need to look at this a bit differently.

We cannot rely on passing events from main. the HuntFinished is to be published only from threre, and it's there only to trigger the report process. this is why we cannot pass events directly to the reporter. 

However, we can use the same way we report vulnerabilities and discovered services, **by collecting them**

1. Creating a new event type `HuntError` for example. and subscribing it in [here](https://github.com/aquasecurity/kube-hunter/blob/473e4fe2b57d3e91b91777de0ed65d13658ae9cb/kube_hunter/modules/report/collector.py#L23)
Follow the same collection process (add a list of `errors`) just above the collector

2. Define `HuntError` [here](https://github.com/aquasecurity/kube-hunter/blob/473e4fe2b57d3e91b91777de0ed65d13658ae9cb/kube_hunter/core/events/types.py#L139)

3. When error occurs, instead of saving the error and propagating, just hit `self.publish_event(HuntError("Error Message"))` from the hunter. because were collecting it (point 1 above) it will end up inside your list

Finally all of the errors could be accessed inside of the reporter (just import the error list [there](https://github.com/aquasecurity/kube-hunter/blob/473e4fe2b57d3e91b91777de0ed65d13658ae9cb/kube_hunter/modules/report/plain.py#L9))

I know I wrote a lot, but it really is will be easier in the future to add more errors that way. and solve our problem.
This is really an important feature to add. I will be more around to help you in the process from now!
Wdyt?

DiptoChakrabarty:
Yeah I am looking through it , it sounds good I am trying to understand it a bit more and then working on it

DiptoChakrabarty:
It is working now I think , getting such output at the end
Selection_159

danielsagi (changes requested):
Hi @DiptoChakrabarty 
Looking at this for the second time, I think we have been looking at this wrongly.
The hunt error basically serves as more advanced logging. (which just spams messages, as the error is triggered for every failed try, which happens alot in a network scan) while we want to make a decision at the end of the hunt, based on what we found.

Want we need is to decide at the end based on already published events, what happend in the scan. sorry to walk through this again from the start, but I tested it myself and the following method works wonderfully

1. Remove the hunt error completely :(
2. Collect `OpenPortEvent` at the collector (like we did last time with the HuntError)
3. At the reporter, just adding logic to the output, something like this:
```python
with open_ports_lock:
    open_ports_len = len(open_ports)
failed_reasons = list()

if not open_ports_len:
    failed_reasons += ["Could not find any open kubernetes ports"]

output += "\nKube Hunter couldn't find any clusters"
if len(failed_reasons):
    output += "\n\nPossible reasons:"
    for i, reason in enumerate(failed_reasons):
        output += f"\n{i+1}. {reason}"
```

DiptoChakrabarty:
> Hi @DiptoChakrabarty Looking at this for the second time, I think we have been looking at this wrongly. The hunt error basically serves as more advanced logging. (which just spams messages, as the error is triggered for every failed try, which happens alot in a network scan) while we want to make a decision at the end of the hunt, based on what we found.
> 
> Want we need is to decide at the end based on already published events, what happend in the scan. sorry to walk through this again from the start, but I tested it myself and the following method works wonderfully
> 
> 1. Remove the hunt error completely :(
> 2. Collect `OpenPortEvent` at the collector (like we did last time with the HuntError)
> 3. At the reporter, just adding logic to the output, something like this:
> 
> ```python
> with open_ports_lock:
>     open_ports_len = len(open_ports)
> failed_reasons = list()
> 
> if not open_ports_len:
>     failed_reasons += ["Could not find any open kubernetes ports"]
> 
> output += "\nKube Hunter couldn't find any clusters"
> if len(failed_reasons):
>     output += "\n\nPossible reasons:"
>     for i, reason in enumerate(failed_reasons):
>         output += f"\n{i+1}. {reason}"
> ```

I think I get the idea , I am working on it , I will publish the OpenPortEvent and catch it in the collector and then report it in a similar fashion in case of error.

DiptoChakrabarty:
Does this look good , made use of the OpenPortEvent only
Selection_167

DiptoChakrabarty:
hey can you please check it out once please

danielsagi:
Hi @DiptoChakrabarty 
I'm sorry I can't sit with you this one out. you probably didn't understand my comments regarding the Error messages.
I meant you should not add published events of open port event.

only add logic on messages at the collector by subscribing.
In your PR you basically added publishing of `OpenPortEvent` s  even on non related port scanning hunters.

I'm closing this PR for now, if you want to address my comments as I stated in a new PR. I could help you again. 
We still would love to see this implemented :)

Issue #502, opened 2022-05-10
kube-hunter v0.6.7: "Configuration is not initialized" when running kube-hunter in container


iElephant:
### What happened

Trying to run the latest docker image (v0.6.7) and get the "Configuration is not initialized" error:

```
docker run --rm -it aquasec/kube-hunter
Traceback (most recent call last):
  File "/usr/local/bin/kube-hunter", line 5, in <module>
    from kube_hunter.__main__ import main
  File "/usr/local/lib/python3.8/site-packages/kube_hunter/__main__.py", line 14, in <module>
    pm = initialize_plugin_manager()
  File "/usr/local/lib/python3.8/site-packages/kube_hunter/plugins/__init__.py", line 16, in initialize_plugin_manager
    pm.load_setuptools_entrypoints("kube_hunter")
  File "/usr/local/lib/python3.8/site-packages/pluggy/_manager.py", line 287, in load_setuptools_entrypoints
    plugin = ep.load()
  File "/usr/local/lib/python3.8/importlib/metadata.py", line 77, in load
    module = import_module(match.group('module'))
  File "/usr/local/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "/usr/local/lib/python3.8/site-packages/kube_hunter_dns_spoof.py", line 4, in <module>
    from kube_hunter.core.events.types import Event, Vulnerability
  File "/usr/local/lib/python3.8/site-packages/kube_hunter/core/__init__.py", line 3, in <module>
    from . import events
  File "/usr/local/lib/python3.8/site-packages/kube_hunter/core/events/__init__.py", line 2, in <module>
    from .handler import EventQueue, handler
  File "/usr/local/lib/python3.8/site-packages/kube_hunter/core/events/handler.py", line 369, in <module>
    config = get_config()
  File "/usr/local/lib/python3.8/site-packages/kube_hunter/conf/__init__.py", line 60, in get_config
    raise ValueError("Configuration is not initialized")
ValueError: Configuration is not initialized

```
The previous version works fine:

```
docker run --rm -it aquasec/kube-hunter:0.6.5
Choose one of the options below:
1. Remote scanning      (scans one or more specific IPs or DNS names)
2. Interface scanning   (scans subnets on all local network interfaces)
3. IP range scanning    (scans a given IP range)
Your choice:
```

### Expected behavior

The kube-hunter should run in docker containers.

[...]

PR #504, opened 2022-05-11
pin image version

rhtenhove:
## Description

This pins the image version in the job manifest. Using this manifest directly (for example  `kubectl apply -f https://raw.githubusercontent.com/aquasecurity/kube-hunter/v0.6.7/job.yaml`) on a pinned version, will allow using that manifest without having to maintain your own, and allowing changes to this repo without breaking a pipeline.

For example, using the job manifest now will fail because of #502 , and there's no previous manifest to refer to.

### Version tag discrepancy

I've also upped the semver tagging in the GitHub actions to coincide with the `kube-bench` format (as well as the format in the tags of this repo). Using the job manifest now will fail, as it first needs to publish with the new tag format.

This shouldn't impact any existing setups, as latest will stay as it is, and older tags will remain as they are.

### Other

Also added label for the pod, and removed backofflimit to just use the default of 6. Cleaner and the same as `kube-bench`.

## Contribution Guidelines

## Fixed Issues

Did not make an issue.

No terminal output changes

## Contribution checklist
 
## Notes

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -39,7 +39,7 @@ jobs:
           password: ${{ secrets.ECR_SECRET_ACCESS_KEY }}
       - name: Get version
         id: get_version
-        uses: crazy-max/ghaction-docker-meta@v1
+        uses: crazy-max/ghaction-docker-meta@v3
         with:
           images: ${{ env.REP }}
           tag-semver: |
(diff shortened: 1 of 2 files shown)


danielsagi:
Hi @rhtenhove, That is a great and neat addition!
Thanks

